PR #38, opened 2018-06-14
Consume API

chriswhong:
This PR disables mirage and has the app consume data from `{api-domain}/projects/:projectid`.

`labs-zap-api` is now serving jsonapi compliant responses on `/projects` that are ready for this branch to consume.

### DIFF
--- a/app/adapters/application.js
+++ b/app/adapters/application.js
@@ -1,5 +1,7 @@
 import DS from 'ember-data';
+import ENV from 'labs-applicant-maps/config/environment';
 const { JSONAPIAdapter } = DS;
+
 
 export default class ApplicationAdapter extends JSONAPIAdapter {
   _ajaxRequest() {
@@ -10,5 +12,5 @@ export default class ApplicationAdapter extends JSONAPIAdapter {
     super._ajaxRequest(...arguments);
   }
 
-  host = 'http://localhost:3000'
+  host = ENV.host;
 }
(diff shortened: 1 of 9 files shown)


app/templates/show-project.hbs
```diff
@@ -16,73 +16,45 @@
 
         <div class="grid-x grid-padding-x">
           <div class="cell medium-6">
+            {{log model}}
```
allthesignals: Plz remove these thanks

app/adapters/application.js
```diff
@@ -9,4 +9,6 @@ export default class ApplicationAdapter extends JSONAPIAdapter {
 
     super._ajaxRequest(...arguments);
   }
+
+  host = 'http://localhost:3000'
```
allthesignals: This shouldn't be hard-coded

config/environment.js
```diff
@@ -44,6 +44,9 @@ module.exports = function(environment) {
     // ENV.APP.LOG_TRANSITIONS = true;
     // ENV.APP.LOG_TRANSITIONS_INTERNAL = true;
     // ENV.APP.LOG_VIEW_LOOKUPS = true;
+    ENV['ember-cli-mirage'] = {
+      enabled: false
```
allthesignals: This should be _enabled_ for development

chriswhong:
All the issues above have been addressed, I added some environment logic to make sure the mirage API calls work in development and the real API gets called in production.

I had added a serializer that works for the real API but not for the mirage results, making this method in the application serializer dependent on environment did the trick.

```
  keyForAttribute(key) {
    return ENV.environment === 'production' ? key : dasherize(key);
  }
```

allthesignals:
💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯

PR #76, opened 2018-06-19
Share Controls

### LINKED ISSUE #71: Build out Share Button

andycochran:
This PR… 
- Adds the `ember-cli-clipboard` addon
- Ports over the share controls and actions from City Map
- Adds a `_m-share-controls` Sass module 
- Changes the project header layout to float the share controls alongside the title

Also… 
- Fixes the project name — `{{#if false}}` was added temporarily for testing and was accidentally committed, and should've been `{{#if model.dcp_projectname}}`

Closes #71.


app/controllers/show-project.js
```diff
@@ -29,4 +35,23 @@ export default class ShowProjectController extends Controller {
...
+  handleShareClose() {
+    this.set('shareClosed', true);
+    this.set('copySuccess', false);
+  }
+
+  @action
+  handleShareSuccess() {
+    this.set('copySuccess', true);
+    run.later(() => {
+      this.set('copySuccess', false);
```
allthesignals: FYI, testing this will random fail for the same reasons we saw earlier which is that async `set`ing on objects is unsafe because they could be destroyed once the callback happens.

PR #96, opened 2018-06-22
Dynamic tiles

chriswhong:
Adds handling for the new `meta` fields in the projects api response:

- `meta.tiles` is the vector tile template for the query, which is used to update the projects map
- `meta.bounds` is the bounds of the project centroids, which is fed to mapboxGL's `map.fitBounds()`


allthesignals:
Boo, failing tests!

PR #143, opened 2018-06-26
nix action status reason filter

andycochran:
This PR temporarily comments out the "Action Status Reason" filter until it's resolved w/ the client how it should work.


allthesignals:
@andycochran need to just figure out conflicts

PR #150, opened 2018-06-27
project map padding

### LINKED ISSUE #149: Add padding to project fitBounds()

andycochran:
This PR adds more padding to the `fitBounds()` on the project map so that more context is visible. Closes #149. 

(Also changes the `.site-name` link to point to the `show-geography` route instead of `application`)


allthesignals:
We should figure out what to do with that application/index route

PR #172, opened 2018-06-29
Stylish Actions 💄💅🏼

andycochran:
This PR… 
- Adds Actions to the Milestones Sass component
- Abstracts the class applying appropriate margin around groups of `.label` elements
- Nixes the possibly-errant pipes ("|") in the Action meta info
- Styles the Action meta in `.label` elements (or "chips" for @chriswhong)

### DIFF
--- a/app/templates/show-project.hbs
+++ b/app/templates/show-project.hbs
@@ -72,9 +72,9 @@
         <div class="grid-x grid-padding-x">
           <div class="cell medium-6">
             <h2>Actions</h2>
-            <ul class="no-bullet">
+            <ul class="actions-list no-bullet">
               {{#each model.actions as |action|}}
-                <li class="grid-x grid-padding-small">
+                <li class="grid-x grid-padding-small action">
                   <div class="cell shrink">
                     <span class="action-icon fa-layers fa-fw">
                       {{fa-icon 'circle' fixedWidth=true class='dark-gray'}}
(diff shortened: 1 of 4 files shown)


app/templates/show-project.hbs
```diff
@@ -72,40 +72,35 @@
...
-                      <span class="medium-gray">|</span>
-                      {{#if action.dcp_ulurpnumber}}
-                        {{#if (eq action.statuscode "Approved")}}
-                          <a href={{build-url "cpcReport" action.dcp_ulurpnumber}} target="_blank">{{action.dcp_ulurpnumber}}</a><sup class="dark-gray">{{icon-tooltip tip='ULURP Number (click on this number to access the City Planning Commision Report).'}}</sup>
+                    <h5 class="no-margin">{{action.dcp_name}}<sup class="dark-gray">{{icon-tooltip tip=(lookup-action-type action.actioncode)}}</sup></h5>
+                    <p class="text-tiny label-group">
+                      <span class="label light-gray">
+                        {{if action.dcp_prefix action.dcp_prefix "No Prefix"}}
+                      </span>{{!-- remove whitespace
+                  --}}<span class="label light-gray">
```
allthesignals: Growing concerned abt code styling here. We have no linters installed for handlebars, but we should start remembering too. My bad!

andycochran: Yeah, I wasn't sure about how to best remove that whitespace. Handlebars whitespace control (`~`) doesn't help us here because of the hardcoded span. If that conditional were block level, it could control the whitespace, but the inline simplicity would be lost.

PR #227, opened 2018-07-04
new tooltips for milestones based on ULURP vs. non-ULURP

trbmcginnis:


app/components/project-milestone.js
```diff
@@ -151,6 +194,24 @@ export default class ProjectMilestoneComponent extends Component {
...
+    const ulurptip = milestoneLookup[milestonename].ulurptooltip;
+    const isUlurp = this.get('isUlurp');
+
+  if (isUlurp) {
+    const  tooltip   = `${ulurptip}` || '';
+    return  tooltip  || `Tooltip not found for milestone ${displayname}`; 
+  } else {
+    const  tooltip  = `${nonulurptip}` || '';
+    return tooltip || `Tooltip not found for milestone ${displayname}`; 
+  }
```
chriswhong: I think we should not have "Tooltip not found" text, every milestone should have one so there's no need for this anymore.  

If there is no tooltip, lets add logic to the handlebars so the tooltip icon doesn't show at all.

allthesignals: Now that that's gone, you can simplify even further like this:
```
return `${nonulurptip}`;
```

PR #228, opened 2018-07-04
12 flag projects

### LINKED ISSUE #12: User can flag projects for data issues
For example, if a link to CPC report is missing

chriswhong:
Adds a pop-up feedback form to the project view.  
The form includes a recaptcha, and sends the project name, project id, and user's feedback to the backend.

Depends on [this PR](https://github.com/NYCPlanning/labs-zap-api/pull/33) for the API changes.

Closes #12

### DIFF
--- a/app/controllers/show-project.js
+++ b/app/controllers/show-project.js
@@ -11,6 +11,8 @@ export default class ShowProjectController extends Controller {
   @argument flagClosed = false;
   @argument copySuccess = false;
 
+  flagText = 'test';
+
   bblFeatureCollectionLayer = {
     "id": "bbl-feature-collection-fill",
     "type": "line",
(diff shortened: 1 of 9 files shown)


app/controllers/show-project.js
```diff
@@ -10,6 +10,11 @@ export default class ShowProjectController extends Controller {
   @argument shareClosed = true;
   @argument copySuccess = false;
 
+  flagText = '';
+  flagClosed = true;
+  flagSuccess = false;
+  reCaptchaResponse = null;
```
allthesignals: In general, unless these things depend on query params (which are inherently Controller level properties), stuff like this should be kept inside a component, including the `bblFeatureCollectionLayer`.

app/controllers/show-project.js
```diff
@@ -72,4 +77,52 @@ export default class ShowProjectController extends Controller {
...
+  }
+
+  @action
+  submitFlag() {
+    const projectname = this.get('model.dcp_projectname');
+    const projectid = this.get('model.dcp_name');
+    const flagText = this.get('flagText');
+    const reCaptchaResponse = this.get('reCaptchaResponse');
+
+    fetch(`http://localhost:3000/projects/feedback`, {
```
allthesignals: This won't work in production... we should use environment variables for this.

app/controllers/show-project.js
chriswhong: Done

app/controllers/show-project.js
chriswhong: Done

allthesignals:
Thanks! 🙌

PR #279, opened 2018-07-11
255 address search

### LINKED ISSUE #255: Address search zoom to map
Address box on map to zoom map to geocoded address

allthesignals:
This closes #255 by plotting a point on the map when user finds an address. Needs design attention from @andycochran


andycochran:
Styled! 

image

PR #285, opened 2018-07-12
change complete to completed

trbmcginnis:


allthesignals:
I think the tests need to be updated

allthesignals:
They're probably checking URLs with "Completed" in them

PR #287, opened 2018-07-12
284 multiple filings

### LINKED ISSUE #284: Filed and Revised Applications
The initial Filed Land Use Application milestone should appear as "Land Use Application Filed" and all later milestones should appear as "Revised Land Use Application Filed"

chriswhong:
Adds logic to change the display values for subsequent application filings.  The first one will show "Land use application filed", subsequent ones will show "Revised Land Use Application Filed"

This is accomplished by adding a computed property in show-project controller that changes the milestonename if there are multiple filings.

Closes #284

### DIFF
--- a/app/templates/show-project.hbs
+++ b/app/templates/show-project.hbs
@@ -111,7 +111,7 @@
             <h2>Milestones</h2>
             <ul class="no-bullet milestones-list">
               {{#if model.milestones}}
-                {{#each model.milestones as | milestone |}}
+                {{#each revisedmilestones as | milestone |}}
                   {{project-milestone milestone=milestone isFiled=isFiled  isUlurp=(eq model.dcp_ulurp_nonulurp "ULURP")}}
                 {{/each}}
               {{else}}
--- a/app/components/project-milestone.js
+++ b/app/components/project-milestone.js
@@ -108,7 +108,16 @@ const milestoneLookup = {
     dateFormat: 'start',
     hideIfFiled: false,
     ulurptooltip: '',
-    nonulurptooltip: '', 
+    nonulurptooltip: '',
+  },
+
+  // special handling for subsequent instances of Filing
+  'Revised Land Use Application Filed Review': {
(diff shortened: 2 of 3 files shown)


app/templates/show-project.hbs
```diff
@@ -111,7 +111,7 @@
             <h2>Milestones</h2>
             <ul class="no-bullet milestones-list">
               {{#if model.milestones}}
-                {{#each model.milestones as | milestone |}}
+                {{#each revisedmilestones as | milestone |}}
```
andycochran: Is `revisedmilestones` a good name for this? They're not all revised.

chriswhong: I am futureproofing, in preparation for another request to change subsequent requests for action X

chriswhong: Or, the array of milestones is revised, even if each milestone is not.

andycochran: It could just be `milestones`. No reason to qualify them.

chriswhong: I wanted to make sure it was different from model.milestones. ` revisedmilestones` reads better in the handlebars, as it's right after a check for `model.milestones`

allthesignals: Agree - this should be computed or something. This should also be wrapped in a component if milestones needs to be munged for presentation - I'm afraid to see the skeletons in the `show-project` controller closet!

chriswhong: It is computed, but it has to happen at the projects level as we already have a component for each milestone.  

We could make a `milestones` component, but that seems like overkill for just one computed, so it may as well stay in `show-project`

andycochran: If the markup is using ~`revised`~`milestones` shouldn't the check be for it, not `model.milestones`?

andycochran: Meh. Semantics. This works. I'm merging it…

Issue #300, opened 2018-07-12
Persist map state, geocoded address, in URL


allthesignals:
When I geocode, click project, then go back, it resets map state

chriswhong:
Jon K has just reported this again.  When user navigates back to the filter view the map should be as they left it.

PR #340, opened 2018-07-13
persist filter view by making show-project a child of projects

### LINKED ISSUE #339: Users want map and filters to stay the same when going back from a profile

chriswhong:
This PR attempts to solve the problem of the filter state resetting when the user navigates back from a project page.

- `show-geography` is renamed to `projects`, and the routes are updated accordingly.
- `show-projects` is rendered into the `{{outlet}}` of `projects`, so when a user clicks on a project in the list view, the route transitions, but the project view is rendered into the existing filter view.  
- Using CSS, we position the projects view over the filter view to keep it hidden from view until the user navigates back.
- When they navigate back, the model has persisted, the list is in the same scroll position the user left it in, the map never went away and doesn't need to initialize again.

Some things to consider:
- I'm not sure of the implications of not clearing the store between these transitions, as the filter and project models share ids for projects, but contain different-ish data.  I haven't seen anything out of the ordinary yet.
- The project view inherits the query params from the filter view... they have no effect on the project view if you load it directly with them there, they just clutter up the URL and are a bit out of place.

TODO: There are inline styles on the containers for `show-project` and `show-project-loading`, these should be classes and styles with SCSS

Closes #339 when complete


allthesignals:
Let’s talk

allthesignals:
> filter state resetting when the user navigates back from a project page

Is this true? I was not seeing this behavior. Does this happen when you click a fresh ‘link-to’ or back button? 

> Using CSS, we position the projects view over the filter view to keep it hidden from view until the user navigates back.

This is not what nested routes are for. Your route nesting should mirror the visual nesting of your UI. Controllers are singletons and their state should not change even when navigating to different routes. I think there are configuration options that make QPs sticky. @andycochran ran into this once.

> clearing the store between these transitions

Hmmm we aren’t clearing the store anywhere anymore 

> filter and project models share ids for projects, but contain different-ish data. I haven't seen anything out of the ordinary yet.

In the show-project route, the ember data ‘findRecord’ API has an option for making another round trip for data so that should not be an issue 

> TODO: There are inline styles on the containers for show-project and show-project-loading, these should be classes and styles with SCSS

Agree!!!

> Closes #339 when complete

Isn’t this also related to #300 in a lot of ways?

chriswhong:
>Is this true? I was not seeing this behavior. Does this happen when you click a fresh ‘link-to’ or back button?

I should not have said "filter state"... I mean to say the component is destroyed, and instantiated again when the user comes back, so they lost all the "load more", their scroll position, and the map state view.

>This is not what nested routes are for. Your route nesting should mirror the visual nesting of your UI. Controllers are singletons and their state should not change even when navigating to different routes. I think there are configuration options that make QPs sticky. @andycochran ran into this once.

I found plenty of examples where people want this functionality in an SPA, for specifically this use case (perusing a list of results, seeing more info about one, and going back to the list to continue perusing), same things that guided its use in the Capital Planning Platform.

allthesignals:
> I found plenty of examples where people want this functionality in an SPA

Gotcha, but this isn’t a point about why the feature should exist but how to approach solving the issue 

> they lost all the "load more", their scroll position, and the map view.

Gotcha. That clarifies a lot. Lets research this more.

chriswhong:
Closing this, as we chose a different direction.

PR #359, opened 2018-07-13
300 persist map state

allthesignals:
This partially addresses #300 by persisting the geocoded address point in the URL and across route transitions. Map state is not yet preserved in URL due to collision with `fitBounds` behavior from the tile server.

This also fixes a bug in which some results were being duplicated in the list. This introduces a controller check to see whether QPs were changed since last route transition.


allthesignals:
Keep this away, it's causing issues I can't track down

PR #404, opened 2018-09-12
[WIP] Change conditions for cpc report link

trbmcginnis:
Update the Model with aftermodel hook to create new boolean property for actions on whether an action should have a CPC report link or not. 

Based on whether the status is "Approved" and what the action is from a list that the Planners gave me.

### DIFF
--- a/app/routes/show-project.js
+++ b/app/routes/show-project.js
@@ -6,6 +6,46 @@ export default class ShowProjectRoute extends Route {
     return this.store.findRecord('project', id, { reload: true });
   }
 
+  afterModel(...args) {
+    const [model] = args;
+
+    model.actions.forEach((cpcAction) => {
+      if (cpcAction.statuscode === 'Approved') {
+        if (cpcAction.actioncode === 'BD'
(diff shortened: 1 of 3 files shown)


trbmcginnis:
This isn't super important! It was just an extra thing that the planners had talked to me about a little while ago. We can definitely revisit later when there's time

app/routes/show-project.js
```diff
@@ -6,6 +6,47 @@ export default class ShowProjectRoute extends Route {
     return this.store.findRecord('project', id, { reload: true });
   }
 
+  afterModel(...args) {
+    const [model] = args;
+
+    model.actions.forEach((cpcAction) => {
+      if (cpcAction.statuscode === 'Approved') {
+        if (cpcAction.actioncode === 'BD'
```
allthesignals: Logic is: if it's approved _and_ the actioncode is one of a list of these? 

Here's one way to do it: 

```
const actionCodes = ['BD', 'HI', 'HK', /* ... more */];
if (cpcAction.statuscode === 'Approved' &&  actionCodes.includes(cpcAction.actioncode)) {
  // other
}
```

PR #456, opened 2019-03-29
Prominent results dates

### LINKED ISSUE #444: Users want to know how the project order in the list is determined
Ordering of search results is opaque

andycochran:
This PR makes the dates in the results list more prominent so that the users knows how the results are sorted. Closes #444. 

image

One thing that is odd: Since we're sorting by reverse chronology of `lastmilestonedate`, and some dates are WAAAY in the future, the most time-sensitive stuff isn't first. We might check with the customer to see if there's a better way we might order the results — e.g. from now forward, followed by from now backward? 

_* Also removes a bunch of crufty CSS._

### DIFF
--- a/app/templates/components/project-list-item.hbs
+++ b/app/templates/components/project-list-item.hbs
@@ -12,15 +12,21 @@
     <span class="label publicstatus-label">{{unless (eq project.dcp_publicstatus_simp 'Unknown') project.dcp_publicstatus_simp "Unknown Status"}}</span>
   </div>
   <div class="cell auto">
-    <h3 class="header-medium no-margin">
+    {{#if project.lastmilestonedate}}
+      <span class="date dark-gray projects-list-result--date">
+        {{icon-tooltip icon='calendar' tip='Latest Milestone'}}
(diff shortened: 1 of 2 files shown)


SomeCityBureaucrat:
Yeah, this is weird, I've noticed something off with this before but never the dates in the future - they are confusing to the public.  An improvement may be having it be only dates prior to today, etc.  Happy to look at any suggestions Labs has for making this better.

app/templates/components/project-list-item.hbs
```diff
@@ -12,15 +12,21 @@
     <span class="label publicstatus-label">{{unless (eq project.dcp_publicstatus_simp 'Unknown') project.dcp_publicstatus_simp "Unknown Status"}}</span>
   </div>
   <div class="cell auto">
-    <h3 class="header-medium no-margin">
+    {{#if project.lastmilestonedate}}
+      <span class="date dark-gray projects-list-result--date">
+        {{icon-tooltip icon='calendar' tip='Latest Milestone'}}
+        {{moment-format project.lastmilestonedate "MMMM D, YYYY"}}
+      </span>
```
allthesignals: Thank you for the new lines!

PR #459, opened 2019-04-01
UI for radius search

### LINKED ISSUE #429: Users want to be able to identify when projects are overlapping on the same site
## Problem
- 46% of all projects share a centroid with another project
- The maximum number of projects sharing a single centroid is 34
- Projects with the same centroid appear as one mapped point; users don’t know if there are multiple projects at a single point 
- In a set of projects sharing a centroid, only the top point is clickable; users can’t click on other projects in that set

_Related problems:_ 
- Users need to see all projects at a specific location or within a specified radius 
- Projects visualized as centroids are more and more misleading as the map is zoomed in, especially once tax lot and building footprints become visible 

## Solution
- [x] Clicking a project centroid on the map no longer navigates directly to that project
- [x] To navigate to a project view, users must select that project from the list 
- When a point is clicked
  - [x] A “filter point” is added to the map
  - [x] Map is zoomed to that point (fitBounds all intersecting projects)
  - [x] A new “Distance from point” filter is enabled 
- When a GeoSearch result is selected
  - [x] A “filter point” is added to the map
  - [x] Map is zoomed to that point (fitBounds all intersecting projects)
  - [x] A new “Distance from point” filter is enabled 
- [x] The “Distance from point” filter restricts the project list results to a specified radius from the user-defined filter point (based on intersecting BBLs, not centroids)
  - [x] Controls allows user to adjust length of radius (slider, text input)
  - [x] Radius defaults to 600 ft
- [x] Subsequent clicks move the filter point 
- [x] Subsequent geosearch searches move the filter point 
- [x] While the “Distance from point” filter is enabled, any map click (not just point clicks) sets/moves the filter point location — this allows the user to see projects within a radius of a specific location that is not a project centroid 
- [x] If “Distance from point” filter is enabled but no filter point is set, filter should show “Click the map to add a point” instead of controls 
- [x] Clearing the search removes the filter point
- [x] Disabling the “Distance from point” filter removes the filter point
- [x] When map is zoomed to a level at which tax lot and building footprints become visible, projects should be visualized as transparent polygons instead of points (styled similar to streets.planning.nyc.gov) which are highlighted when their corresponding list item is hovered

Enabled “Distance from point” filter (with and w/o a filter point): 

image

image

Filter point set to zero radius: 

image

Filter point set to larger radius: 
image

image

allthesignals:
Initial PR that adds the UI and appropriate API querying for clicking, searching by distance from point. Depens on API chaanges for filter to work

ntoe: tests fail for innocuous reasons, will fix soon

Closes #429


andycochran:
- [ ] Clicking the filter toggle button throws an error (clicking the filter name toggles the filter correctly): `Assertion Failed: <labs-zap-search@component:filte…r286> had no action handler for: mutateWithAction`

- [ ] If you click a point while zoomed out to enable the distance filter, then zoom in, the distance filter no longer lines up with the point (Is this because we use the map click lat/lon to place the filter instead of the actual lat/lon of the feature point?) 
  image

- [ ] The pink distance circle should be sized to match the radius set in the slider. Currently it's an explicit pixel size that's the same at all zoom levels. However, the circle should have a minimum size so it's still visible when the radius set to is 0.

allthesignals:
> If you click a point while zoomed out to enable the distance filter, then zoom in, the distance filter no longer lines up with the point (Is this because we use the map click lat/lon to place the filter instead of the actual lat/lon of the feature point?)

That's exactly why!

andycochran:
- [ ] Sometimes the input does not update when the slider changes

- [ ] Label is rounded but actual value isn't, which causes the input to get set to 599.9999 when the slider is set to 600 (e.g.)

andycochran:
- [ ] When the map zooms while the distance filter is enabled, the fitBounds() should include the dotted circle (so the user will always sees the full radius and know what geographic area is being filtered). _Currently, if a large radius only has one project, you zoom way in and don't see the dotted circle._

andycochran:
- [ ] When the distance filter is enabled, clicking anywhere on the map should move the `distance_from_point` lat lon to that clicked location.

allthesignals:
> When the map zooms while the distance filter is enabled, the fitBounds() should include the dotted circle (so the user will always sees the full radius and know what geographic area is being filtered). Currently, if a large radius only has one project, you zoom way in and don't see the dotted circle.

@andycochran how badly do we need this? Since it's pretty idiosyncratic (it will always fit bounds to the result set), it'll be a heavier lift than we like...

.eslintignore
```diff
@@ -0,0 +1,20 @@
...
+/tmp/
+
+# dependencies
+/bower_components/
+/node_modules/
+
+# misc
+/coverage/
+!.*
+
```
allthesignals: test

app/components/filter-distance-from-point.js
```diff
@@ -0,0 +1,58 @@
...
+  @action
+  handleClick(e) {
+    const { target: map } = e;
+    const { point } = e;
+    const [feature] = map.queryRenderedFeatures(
+      point,
+      { layers: [this.pointLayerId] },
+    );
+
+    if (feature) {
```
trbmcginnis: I console logged `if (feature)` vs. `else if (this.shouldQueryFullMap)` and I couldn't find a case when `if (feature)` was true. I'm assuming this is checking whether a feature point for the radius already exists on the map, so not sure why this wouldn't be true if I had made multiple points. Could you explain a little more what this does/is checking for?

app/components/address-search.js
```diff
@@ -0,0 +1,66 @@
...
+    'circle-opacity': 0,
+    'circle-stroke-opacity': 0.2,
+  },
+};
+
+export default class AddressSearch extends Component {
+  // @argument
+  map;
+
+  // @argument
```
trbmcginnis: Can you add comments for these to explain the expected behavior?

app/components/address-search.js
```diff
@@ -0,0 +1,66 @@
...
+  onSelectSearchResult = () => {}
```
trbmcginnis: Can you add comments for `onSelectSearchResult` and `onClearSearchResult` to explain their expected behavior?

app/components/filter-distance-from-point.js
allthesignals: @trbmcginnis did you try logged it when the filter is off and you click on a point?

app/components/address-search.js
allthesignals: @trbmcginnis sure, full comments forthcoming. The front end doesn't do any of the filtering, the backend does. All the front end is responsible for is sending point coordinates and radius distance to the backend, and providing visual feedback to users.

app/controllers/query-parameters/show-geography.js
```diff
@@ -98,6 +98,20 @@ export const projectParams = new QueryParams({
       return value.split(',').sort();
     },
   },
+  distance_from_point: {
+    defaultValue: [0, 0],
```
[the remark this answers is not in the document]
allthesignals: > Is there a reason why we want a default point?

Yep! It expects a valid coordinate

PR #480, opened 2019-04-16
Add documentation for new and existing components

allthesignals:
This documents a large # of classes and methods, and hopefully helps with navigating the source code.

### DIFF
--- a/app/components/filter-distance-from-point.js
+++ b/app/components/filter-distance-from-point.js
@@ -1,31 +1,57 @@
 import Component from '@ember/component';
-// import { argument } from '@ember-decorators/argument';
 import { action, computed } from '@ember-decorators/object';
 import { generateCircleFromFeet } from 'labs-zap-search/helpers/generate-circle-from-feet';
 
-// it takes a point and radius, computes a geojson circle
-// and handles clicks that query mapbox-gl for intersections
-// triggers optional events
+/**
+ * The FilterDistanceFromPoint component takes a point and radius, computes a geojson circle
+ * and handles click events that query mapbox-gl for intersections triggers optional events.
+ */
 export default class FilterDistanceFromPoint extends Component {
-  // @argument
+  /**
+   * {{ember-mapbox-gl}} contextual component hash, used in template. See https://github.com/kturney/ember-mapbox-gl.
+   * @argument
+   * @required
+   */
   map;
 
-  // @argument
+  /**
+   * Point coordinates for filter query
+   * @argument{Array}
+   * @required
+   */
   pointGeometry;
 
-  // @argument
+  /**
+   * Radius from point in feet
+   * @argument
+   * @required
+   */
   radius;
 
-  // @argument
+  /**
+   * Event triggered when map is clicked for the purpose of radius filter;
+   * @argument
+   * @public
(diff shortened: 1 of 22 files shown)


app/components/filter-distance-from-point.js
```diff
@@ -1,41 +1,69 @@
...
-// import { argument } from '@ember-decorators/argument';
 import { action, computed } from '@ember-decorators/object';
 import { generateCircleFromFeet } from 'labs-zap-search/helpers/generate-circle-from-feet';
 
-// it takes a point and radius, computes a geojson circle
-// and handles clicks that query mapbox-gl for intersections
-// triggers optional events
+/**
+ * The FilterDistanceFromPoint component takes a point and radius, computes a geojson circle
+ * and handles click events that query mapbox-gl for intersections triggers optional events.
```
julialucyhogan: im a bit confused by the language here. want to make sure I understand:  It handles click events (on what?), and defines the actions for those click events that query mapbox-gl for intersections with the computed circle, and then triggers optional events

app/components/filter-distance-from-point.js
```diff
@@ -1,41 +1,69 @@
...
 
     return generateCircleFromFeet([pointGeometry, radius]);
   }
 
-  // queries relevant layer for intersecting feature and sends it
-  // to the click action. conditionally allows for clicking
-  // anywhere on the map
+  /**
+   * Click handler action handed to mapbox-gl; when triggered,
+   * it queries the specified point layer for an intersecting feature and sends it
```
julialucyhogan: can we pick a better word for specified? What is the point layer that we're talking about ? how do we pick the point layer?

app/components/structural/project-list-map.js
```diff
@@ -72,25 +71,44 @@ export const projectPolygonsHoverLayer = {
...
-// and is very much on the domain-problem side of the spectrum
-// it also yields out some contextual components that are invoked
-// from the consuming end. it also threads down an action to determine
-// that tile mode.
-
-// tile mode is what determines whether we see polygon or centroids.
+/**
+  * This component renders a map and receives a tiles URL
+  * for rendering dynamic tiles. It's responsible for some styling
+  * and is very much on the domain-problem side of the spectrum
```
julialucyhogan: i am a n00b; can you explain what you mean by "domain-problem" and "consuming end"?

app/components/filter-distance-from-point.js
allthesignals: Handles click events on the mapbox-gl map itself. Yes. gonna revise for clarity!

app/components/filter-distance-from-point.js
allthesignals: Yes, the point layer here is a mapbox-gl layer, determined by the `pointLayerId` property. It's actually a constant here, I made it a class field for no particular reason. I think I will make it a constant so it's clear that this is something static.

app/components/structural/project-list-map.js
allthesignals: Hmm I mean that it's something structural as opposed to generic? The "consuming end" means where the component is invoked. 

Will change language for clarity1

PR #505, opened 2019-04-24
List applied filters when there are no results

allthesignals:
Use ember-intl to “translate” between column names and person-friendly titles, keeping such mappings in one place without overhauling the overall architecture of the show-geography controller.

### DIFF
--- a/app/components/applied-filters-list.js
+++ b/app/components/applied-filters-list.js
@@ -1,0 +1,26 @@
+import Component from '@ember/component';
+import { computed } from '@ember-decorators/object';
+import { getLabelFor } from 'labs-zap-search/helpers/get-label-for';
+
+/**
+ * Component that takes an array, runs it through a service method
+ * for translations, and prints them on the page with a friendly msg.
+ */
+export default class AppliedFiltersListComponent extends Component {
+  /**
+   * Array of column names, will be passed through a translator
+   *
+   * @argument(Array)
+   */
+  appliedFilters;
+
+  /**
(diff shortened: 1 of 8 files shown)


app/templates/show-geography.hbs
```diff
@@ -99,7 +99,7 @@
         filterNames='project_applicant_text'
         as |section|}}
         {{#section.filter-wrapper
-          filterTitle='Text Match'
+          filterTitle=(t 'filters.project_applicant_text')
```
trbmcginnis: Ooh love how these translated titles are now reusable throughout the app! :)

julialucyhogan:
Since we have no intentions (i dont think) of actually translating to multiple languages, I'm flopping back and forth btwn if I think this is the way to go, or if it would be better to write a little helper that just does the look up. Since the logic is super simple on the one hand it feels like a lot of extra configuration and an external dependency that's not really necessary. On the other hand, the lookup logic is handled for us. Just thought it was worth starting a convo about. Thoughts?

app/components/applied-filters-list.js
```diff
@@ -0,0 +1,32 @@
...
+  @service
+  intl;
+
+  /**
+   * Unique "humanized" filter names from the translations/en-us.yaml file.
+   */
+  @computed('appliedFilters')
+  get currentFiltersNames() {
+    return this.appliedFilters
+      .map(filterName => this.intl.t(`filters.${filterName}`))
```
julialucyhogan: this should use get-label-for, right ?

PR #506, opened 2019-04-24
478: Add test selectors to optimize filter tests

### LINKED ISSUE #478: Make filter tests less brittle
- [ ] acceptance/filter-checkbox-test.js needs better IDs for each of the filter toggles for testing purposes https://github.com/NYCPlanning/labs-zap-search/blob/develop/tests/acceptance/filter-checkbox-test.js#L78

trbmcginnis:
Add in `ember-test-selectors` so that we can optimize `filter-checkbox-test` and remove the need to select by `li: first-child`

**I also noticed that we had a lot of empty integration tests, I can jump on those if we think it's necessary**

Closes #478

### DIFF
--- a/app/templates/components/filter-wrapper.hbs
+++ b/app/templates/components/filter-wrapper.hbs
@@ -1,4 +1,7 @@
-<div class="filter-header">
+<div
+  class="filter-header"
+  data-test-filter-section={{dasherizedFilterTitle}}
+>
   <div class="switch tiny float-left">
     <input class="switch-input" type="checkbox" checked={{if (contains filterNames appliedFilters) 'checked'}}>
     <label {{action mutateWithAction}} class="switch-paddle">
@@ -12,6 +15,7 @@
   </h6>
 </div>
 
-<div class="filter-controls">
+<div class="filter-controls"
+ data-test-filter-control={{dasherizedFilterTitle}}>
   {{yield}}
 </div>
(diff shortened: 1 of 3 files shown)


app/templates/components/filter-wrapper.hbs
```diff
@@ -1,4 +1,7 @@
-<div class="filter-header">
+<div
+  class="filter-header"
+  data-test-filter-section={{dasherizedFilterTitle}}
```
trbmcginnis: @allthesignals What benefit is there to using a test selector here rather than just the actual class? (e.g. `.filter-section-community-district`)?

allthesignals: @trbmcginnis good Q, the benefits are that these test-specific selectors are stripped out of the production build of the app. Using a regular class selector is fine too, until @andycochran needs to update or remove those classes entirely. Yeah, this would cause the test suite to fail and be caught, but it makes a CSS selector responsible for testing and styling all at once.

tests/acceptance/filter-checkbox-test.js
```diff
@@ -33,16 +34,16 @@ module('Acceptance | filter checkbox', function(hooks) {
     server.createList('project', 20);
     await visit('/');
     await click('.filter-section-community-district .ember-power-select-multiple-options');
-    await fillIn('.filter-section-community-district .ember-power-select-multiple-options input', 'Brooklyn 1');
-    await click('.ember-power-select-options li:first-child');
+    await fillIn('.filter-section-community-district .ember-power-select-multiple-options input', 'Brooklyn');
+    await selectChoose('.community-district-dropdown-selection', 'Brooklyn 2');
```
allthesignals: COOL! Did not know that existed! Nicely done.

app/templates/components/filter-wrapper.hbs
trbmcginnis: Okay cool I'm going to go ahead and change those selectors in the test and re-commit, thanks!

PR #507, opened 2019-04-29
Upgrade nouislider; remove date “round” to nearest year

### LINKED ISSUE #360: ZAP front page - slider issue
Hi all. I found a ZAP issue on the front page. For the sliders, I can't get the front slider to land on the year of 1980. The second slide works fine, but I can't set it as a start date. It jumps to 1979 or 1981. I tried on Rachaele's PC and she had the same issue. Thanks!

allthesignals:
Previously, the date-ranger-filter was rounding epoch time back to the _end_ of the previous year instead of the very beginning of the input yet. It's unclear why but it was off by one.

The fix was to add this offset into the commit.

Closes #360

### DIFF
--- a/app/components/date-range-filter.js
+++ b/app/components/date-range-filter.js
@@ -53,15 +53,25 @@ export default class SliderFilterComponent extends Component {
   replaceProperty = () => {}
 
   /**
-   * Passed to {{range-slider}}; triggered when {{range-slider}} mutates some data
+   * Callback that gets passed to {{range-slider}}; triggered when {{range-slider}}
+   * mutates some data. It receives numerics from range-slider, they are sometimes
+   * not rounded. This callback performs some cleaning by rounding the values into
+   * integers and "rounding" the epoch timestamps to be inclusive of the full year
+   * they intersect with.
+   *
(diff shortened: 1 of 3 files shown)


app/components/date-range-filter.js
```diff
@@ -58,11 +58,11 @@ export default class SliderFilterComponent extends Component {
    */
   @action
   sliderChanged([min, max]) {
-    // because the slider returns unix epochs based on its own step increment,
-    // get the startOf() the min timestamp's year, and the endOf() of the max timestamp's year
-    const minStart = parseInt(moment(min, 'X').utc().startOf('year').format('X'), 10);
-    const maxEnd = parseInt(moment(max, 'X').utc().endOf('year').format('X'), 10);
+    // must round integers because range slider sometimes provides
+    // decimal values in the callback
```
julialucyhogan: :(

allthesignals:
I need to add the rounding back in but offset it by one so it doesn't fall into the previous year

app/components/date-range-filter.js
allthesignals: It may not matter now but I want to control the rounding so it's predictable. I don't know why nouislider is doing this.

PR #514, opened 2019-05-07
513: add CEQR number tag to project list

### LINKED ISSUE #513: Add CEQR number to projects list
Add CEQR number as a tag to the projects list

trbmcginnis:
- Add `dcp_ceqrnumber` and `dcp_ceqrtype` as a single tag to projects list. 
- If both `dcp_ceqrtype` and `dcp_ceqrnumber` are NULL, the tag does not show up. 
- If `dcp_ceqrtype` is "Unlisted" and `dcp_ceqrnumber` is NULL, the tag does not show up. 
- Some projects have an associated CEQR type but no associated number, and vice versa. In this case, the tag still shows up, but only the available info is shown. 

Closes #513


trbmcginnis:
Screen Shot 2019-05-07 at 12 01 04 PM

PR #523, opened 2019-05-22
521: update logic for displaying BBLs on map on project view

### LINKED ISSUE #521: Projects that aren't mapped have blank maps on their profile
This is a problem because it zooms to a default location which could be misleading. It also makes it seem like the map is buggy or there's a data quality issue. The map should be removed and/or the profile should clearly indicate that the project doesn't have spatial data.

Examples:
- 2019Q0306
- P2017Q0185

trbmcginnis:
I think there might have been a data format change which caused a problem for the original logic that we were using to display BBLs on a project map vs. display "No Map Available". Our original logic was `#if model.bbl_featurecollection` in `show-project.hbs`. 

This PR changes `model.bbl_featurecollection` to `model.bbls` to check for array of BBLs. 

Closes #521

### DIFF
--- a/app/templates/show-project.hbs
+++ b/app/templates/show-project.hbs
@@ -154,7 +154,7 @@
 
       </div>
       <div class="cell large-5">
-        {{#if model.bbl_featurecollection}}
+        {{#if model.bbls}}
           {{#mapbox-gl
               id='project-map'
               initOptions=(hash


app/controllers/show-project.js
```diff
@@ -61,6 +61,17 @@ export default class ShowProjectController extends Controller {
     });
   }
 
+  // checks that array of BBLs exists, and if so returns true
+  // this is used to either display the BBLs for a project page on the map (true) OR
+  // to display "No Map Available" when a project has no BBLs (false)
+  @computed('model.bbl_featurecollection')
+  get hasBBLs() {
+    const { bbls } = this.model;
```
julialucyhogan: why can't we just check that model.bbls exists instead of model.bbl_featurecollection in the template?

allthesignals: Also I'd add that the dependent key for this computed property is never used here. In theory you could just `bbls` directly on the model but it wouldn't recomputed (I know we'd never reasonably do that but in general). I'd say that `model.bbls` is the dependent key since that's what's being used to determine whether or not it has bbls, no?

trbmcginnis: u right

app/templates/show-project.hbs
```diff
@@ -154,7 +154,7 @@
 
       </div>
       <div class="cell large-5">
-        {{#if model.bbl_featurecollection}}
+        {{#if model.bbls}}
```
allthesignals: What is the data type of `bbls`? Any array? Should be fine in the template (which considers empty array falsy).

trbmcginnis: `model.bbls` is an array, yes. On cases when there are no bbls, `model.bbls` just comes back as `null`

PR #527, opened 2019-05-29
526: No Map Available should show up if project does not have bbl geometries

### LINKED ISSUE #526: Projects WITH bbls but WITHOUT bbl geometry should have "No Map Available"
Currently, "No Map Available" is based on whether a project has a list of bbls `model.bbls` rather than if a project has bbl geometries `this.model.bbl_featurecollection.features[0].geometry`. This should be changed to check for geometry, not the list.

trbmcginnis:
I originally assumed when fixing this bug that all projects with a list of BBLs would have associated geometry, but some projects might have a list of BBLs but do NOT have associated BBL geometry (for example, older projects).

Currently, "No Map Available" is based on whether a project has a list of bbls `model.bbls` rather than if a project has bbl geometries `this.model.bbl_featurecollection.features[0].geometry`. This PR just edits it to check for geometry AND to check if a feature collection exists at all `this.model.bbl_featurecollection.features.length`, instead of checking for a list.

This PR also adds a default value to the `bbl_featurecollection` attribute in the projects model. 

Closes #526

### DIFF
--- a/app/controllers/show-project.js
+++ b/app/controllers/show-project.js
@@ -61,6 +61,12 @@ export default class ShowProjectController extends Controller {
     });
   }
 
+  @computed('model.bbl_featurecollection')
+  get hasBBLFeatureCollectionGeometry() {
+    return this.model.bbl_featurecollection.features.length
+    && this.model.bbl_featurecollection.features[0].geometry;
+  }
+
   @action
   handleMapLoad(map) { // eslint-disable-line
     window.map = map;
--- a/tests/unit/controllers/show-project-test.js
+++ b/tests/unit/controllers/show-project-test.js
@@ -1,10 +1,11 @@
 import { module, test } from 'qunit';
 import { setupTest } from 'ember-qunit';
+import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
 
 module('Unit | Controller | show-project', function(hooks) {
   setupTest(hooks);
+  setupMirage(hooks);
 
-  // Replace this with your real tests.
   test('it runs handleMapLoad', function(assert) {
     const controller = this.owner.lookup('controller:show-project');
     assert.ok(controller);
@@ -26,5 +27,44 @@ module('Unit | Controller | show-project', function(hooks) {
       addControl() {},
       fitBounds() {},
     });
+  });
+
+  test('hasBBLFeatureCollectionGeometry is null when geometry is null', async function(assert) {
+    const controller = this.owner.lookup('controller:show-project');
+    assert.ok(controller);
+
+    const project = server.create('project', {
+      bbl_featurecollection: {
+        features: [{
+          type: 'Feature',
+          geometry: null,
+        }],
+      },
+    });
+    const projectModel = await this.owner.lookup('service:store').findRecord('project', project.id);
+
+    controller.model = projectModel;
+
+    const hasNoGeometry = controller.hasBBLFeatureCollectionGeometry;
+
+    assert.equal(hasNoGeometry, null);
+  });
+
+  test('hasBBLFeatureCollectionGeometry is 0 when features is empty', async function(assert) {
+    const controller = this.owner.lookup('controller:show-project');
+    assert.ok(controller);
+
+    const project = server.create('project', {
+      bbl_featurecollection: {
+        features: [],
+      },
+    });
+    const projectModel = await this.owner.lookup('service:store').findRecord('project', project.id);
+
+    controller.model = projectModel;
+
+    const hasEmptyFeatures = controller.hasBBLFeatureCollectionGeometry;
+
+    assert.equal(hasEmptyFeatures, 0);
   });
 });
--- a/app/models/project.js
+++ b/app/models/project.js
@@ -4,6 +4,17 @@ import { computed } from '@ember-decorators/object';
 
 const { Model } = DS;
 
+const EmptyFeatureCollection = {
+  type: 'FeatureCollection',
+  features: [{
+    type: 'Feature',
+    geometry: null,
+    properties: {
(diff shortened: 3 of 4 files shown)


app/controllers/show-project.js
```diff
@@ -61,6 +61,14 @@ export default class ShowProjectController extends Controller {
     });
   }
 
+  @computed('model.bbl_featurecollection')
+  get hasBBLFeatureCollectionGeometry() {
+    const geom = this.model.bbl_featurecollection.features[0].geometry;
```
allthesignals: What does it look like in cases where there is no bbl geom? I ask because a feature collection can be empty:

```javascript
{
  type: 'FeatureCollection',
  features: [],
}
```

In that case, you would get an unhandled exception: "cannot find property 'geometry' of undefined'". 

If we look at the model attr for bbl_featurecollection, there's no default: https://github.com/NYCPlanning/labs-zap-search/blob/develop/app/models/project.js#L78. 

So we can't guarantee it to be not null. Thoughts? 

I think if it were me, I would do it like this:

```javascript
@computed('model.bbl_featurecollection')
get hasBBLFeatureCollectionGeometry() {
  return this.model.bbl_featurecollection.features.length;
}
```

This could return 0, which evaluates to "falsey" in the template, or some number greater than 0, which will be "truthy" in the template.

Then, I would go here to set up a default so we can guarantee that the model is always some kind of feature collection:
https://github.com/NYCPlanning/labs-zap-search/blob/develop/app/models/project.js#L78

You can see an example of this here: https://github.com/NYCPlanning/labs-applicantmaps/blob/develop/app/models/geometric-property.js#L107-L108.

Happy to pair on this!

trbmcginnis: Awesome, thank you! Confused about how this change covers the case where a project DOES have a feature collection BUT the geometry is `null`? Because in the case that I'm thinking of, a feature collection exists but only the geometry is `null`, meaning the computed above would not cover that situation, because the `length` would still come back as greater than 0. 

This is currently happening in the new Gowanus project, because BBL geometry has not been loaded yet for this project, but it DOES have a feature collection. <--- this might be a bug, so maybe this situation should never happen, and if that's the case then `this.model.bbl_featurecollection.features` would only exist for projects that for sure had `geometry` not equal to `null`

allthesignals: @trbmcginnis gotcha, I see! A null geometry is also definitely compliant with the GeoJSON spec, too.

In that case, I would add your logic back! -> 

```javascript
return this.model.bbl_featurecollection.features.length 
  && this.model.bbl_featurecollection.features[0].geometry;
```

I'm not sure why but you need the length check before you check for the geometry to avoid situations with empty features. So, I still think we need a default value for the model. Thoughts?

trbmcginnis: Sweeettt, I'll go ahead and make that change :).

Yes, I agree that adding a default value to the model is a good idea--anything to make it more secure

allthesignals:
@trbmcginnis sick! Thank you thank you. Could we add a test somewhere? Happy to work together. I just think that when we find bugs, we should at least add tests for those bugs. Sorry I keep asking for more! Bounce it over my way if you're busy.

julialucyhogan:
Looks good, but I'm going to second Matt's ask for tests. Altho I'm going to challenge his invite to bounce it over his way. I am guilty of taking him up on that offer, but I don't think it's how we progress to become a well oiled test writing machine!! We all need to learn to be comfortable writing tests. Pairing (sorry matt :) ) is prob the best way forward here for now.

trbmcginnis:
> Looks good, but I'm going to second Matt's ask for tests. Altho I'm going to challenge his invite to bounce it over his way. I am guilty of taking him up on that offer, but I don't think it's how we progress to become a well oiled test writing machine!! We all need to learn to be comfortable writing tests. Pairing (sorry matt :) ) is prob the best way forward here for now.

I think that's a great idea--I'll write a test for this today/tomorrow before merge!!

tests/unit/controllers/show-project-test.js
```diff
@@ -27,4 +26,39 @@ module('Unit | Controller | show-project', function(hooks) {
...
+  });
+
+  test('hasBBLFeatureCollectionGeometry is null when features is empty', function(assert) {
+    const controller = this.owner.lookup('controller:show-project');
+    assert.ok(controller);
+
+
+    controller.model = {
+      bbl_featurecollection: {
+        features: [{}],
```
allthesignals: This is slightly different from `features: []`. In JavaScript, an empty object evaluates to truthy. So, I think you're not quite testing for the behavior we're talking about, which is "features is empty". Thoughts?

tests/unit/controllers/show-project-test.js
```diff
@@ -27,4 +26,39 @@ module('Unit | Controller | show-project', function(hooks) {
...
+
+
+    controller.model = {
+      bbl_featurecollection: {
+        features: [{
+          type: 'Feature',
+          geometry: null,
+        }],
+      },
+    };
```
allthesignals: I woulda preferred using mirage to set a real model on the controller, but the stub should do for now.

trbmcginnis: Okay I'll try to set one up

tests/unit/controllers/show-project-test.js
trbmcginnis: Changed!

tests/unit/controllers/show-project-test.js
trbmcginnis: Changed!

app/controllers/show-project.js
trbmcginnis: Changed!

Issue #548, opened 2019-07-24
Add models and create Mirage data


hannahkates:
**Models:**
- [ ] **User** This will come from the `dcp_projectlupteam` table.
  - [ ] Name
  - [ ] Type of affiliation (community board, borough board, or borough president). This will determine which type of disposition/recommendation form they get.
  - [ ] Affiliation (which specific community board or borough prez office? ex: BX5)
  - [ ] Title (district manager or other)
  - [ ] Array of projects they're assigned to
- [ ] **Project**. We want all the attributes currently returned on the `zap-api/project/:id` route.
- [ ] **Action**. Multiple actions per project. One hearing date per action and potentially multiple recommendations
- [ ] **User's hearing submission** (submitted before the recommendation) See #547 for the field names in ZAP CRM.
  - [ ] Public Hearing Scheduled (Y/N)
  - [ ] Public Hearing Scheduled Location (text)
  - [ ] Date of Public Hearing (date and time)
- [ ] **User's disposition/recommendation submission**. Necessary attributes vary depending on the user's type of affiliation and the project. See screenshots below for the three types. See #547 for the field names in ZAP CRM.
  - [ ] Borough President
  - [ ] Borough Board (submitted by Borough President user, not Community Board)
  - [ ] Community Board

[**ZAP CRM metadata for reference**](https://drive.google.com/file/d/1jLx4YL2gyhBE20zG9mTLFiTTxdQb0YHL/view?usp=sharing)

**Screenshots:**

![image](https://user-images.githubusercontent.com/13967925/61964832-dc762500-af9c-11e9-90d5-688dfa60d6cb.png)

![image](https://user-images.githubusercontent.com/13967925/61965055-59a19a00-af9d-11e9-8690-63e09a467888.png)

Borough President
![image](https://user-images.githubusercontent.com/13967925/61965141-881f7500-af9d-11e9-8fca-74a55d55e092.png)

Borough Board
![image](https://user-images.githubusercontent.com/13967925/61965088-6f16c400-af9d-11e9-80d1-665bf75b10f3.png)

Community Board
![image](https://user-images.githubusercontent.com/13967925/61965040-4c84ab00-af9d-11e9-92d4-c77ea7d7c5c3.png)

hannahkates:
General Information (2 column layout) 

Disposition Name: Read only, Auto populated 

Project: Read only, Auto populated 

Project Name: Read only, Auto populated 

Project Action: Read only, Auto populated 

[...]

PR #557, opened 2019-07-31
[WIP] 548: create models and mirage for LUPP refactor

godfreyyeung:
- First pass at creating new Ember models for the LUPP refactor
- new models include `user`, `action`, `hearing`, `recommendation`, and `user-project-participation`
- accompanying mirage data will be created in another branch

Comments throughout provide an overview of how frontend models/fields eventually map to CRM models/fields 

You can comment on the model relationship [visualization](https://miro.com/welcomeonboard/AFpxIbORLIoIM5UF2pdnKcpBgYRYe8JFFkejggKEskW9B9eNl0BvlzCQ3LXJg9oA).

Addresses #548


app/models/action.js
```diff
@@ -0,0 +1,44 @@
...
+// DB table: dcp_projectaction
+
+  // One Project to Many Actions
+  @belongsTo('project') project;
+
+  // One Hearing to One Action
+  @belongsTo('hearing') hearing;
+
+  // Many Recommendations to One Action
+  @hasMany('recommendation') recommendations;
```
hannahkates: This should be the other way around. One recommendation can have many actions. Participants have the option of either A) submitting multiple recommendations, one recommendation for each action, or B) submitting a single recommendation for all actions at once.

app/models/hearing.js
```diff
@@ -0,0 +1,20 @@
+import DS from 'ember-data';
+import { attr, belongsTo } from '@ember-decorators/data';
+
+const { Model } = DS;
+
+export default class HearingModel extends Model {
+// DB Table: dcp_communityboarddisposition
+
+  // One Action to One Hearing
+  @belongsTo('action') action;
```
hannahkates: Hearing belongs to project, not action. LUPs only schedule one hearing for a project, regardless of how many actions it has.

app/models/hearing.js
```diff
@@ -0,0 +1,20 @@
...
+
+  // calculated based on date
+  @attr('boolean') isScheduled;
```
hannahkates: Where will the calculation happen? In the form controller? I thought it could be done here in the model but maybe not.

app/models/recommendation.js
```diff
@@ -0,0 +1,45 @@
+import DS from 'ember-data';
+import { attr, belongsTo } from '@ember-decorators/data';
+
+const { Model } = DS;
+
+export default class RecommendationModel extends Model {
+// DB table: dcp_communityboarddisposition
+
+  @belongsTo('user') user;
```
hannahkates: I think it belongs to action, not user. One user can have multiple recommendations, one recommendation per action.

app/models/recommendation/borough-board.js
```diff
@@ -0,0 +1,7 @@
+import CommunityBoardModel from './community-board';
+
+export default class RecommendationBoroughBoardModel extends CommunityBoardModel {
+// extension of CommunityBoardModel with no changes
+// CommunityBoardModel is an extension of RecommendationModel with added attributes
+
```
hannahkates: This is missing the necessary vote attributes. It should look similar (if not the same as?) the Community Board one

app/models/user.js
```diff
@@ -0,0 +1,34 @@
...
+const { Model } = DS;
+
+export default class UserModel extends Model {
+// DB table: dcp_projectlupteam
+
+  // user's internal id should be their email
+
+  // One Project Has Many Users
+  @hasMany('project') projects;
+
```
hannahkates: Users can also have many projects. Is that adequately captured as is?

app/models/hearing.js
trbmcginnis: In the spreadsheet, it mentioned it was "calculted based on date"--I'm assuming this means the public hearing date?

app/models/recommendation/borough-board.js
trbmcginnis: This is an extension of the Community Board model so it will already have all of the appropriate attributes included int he Community Board one

app/models/action.js
allthesignals: So are these many-to–many? Many recommendations can be related to many actions?

trbmcginnis: Ahh okay gotcha --will make the changes accordingly

allthesignals:
I would definitely start adding in some mirage setup as soon as you can because that will begin to inform whether your associations are set up correctly, which can be a little tough.

The jury's out on whether or not you should test model relationships. See here: https://guides.emberjs.com/release/testing/testing-models/#toc_testing-relationships

> Ember Data contains extensive tests around the functionality of relationships, so you probably don't need to duplicate those tests. You could look at the Ember Data tests for examples of deeper relationship testing if you feel the need to do it.

app/models/hearing.js
trbmcginnis: Okay cool so the relationship is: ONE Project to ONE Hearing--gotcha!

app/models/recommendation.js
trbmcginnis: Okay so the relationships are: 1) One Recommendation to One Action **_OR_** 2) One Recommendation to Many Actions

app/models/user.js
trbmcginnis: Sorry the comment seems to be out of order but the relationship is coded correctly. Will alter

godfreyyeung: Should be captured. Actually, we may even get rid of the User model's `@hasMany('project') projects;` and the Project model's `@hasMany('user') users;`, since those associated can be acquired through `userProjectParticipation`

trbmcginnis:
> I would definitely start adding in some mirage setup as soon as you can because that will begin to inform whether your associations are set up correctly, which can be a little tough.

Yep! That's being done in another branch once the team has approved the data relationships are good to go

app/models/action.js
trbmcginnis: ohhh yes it's many to many!

PR #564, opened 2019-08-06
Add simple auth with implicit grant configured; Add test for sign-in

allthesignals:
This PR adds authentication via NYC.ID.

NOTE: when you try to login, it will redirect you to production. In order to test that it works, you must change the URL from 'zap.planning.nyc.gov' to whatever the generated Netlify preview domain is.

The main event is the application route. This takes a mixin which looks for the `access_token` access, per the OAuth implicit grant standard.


app/components/sign-in.js
```diff
@@ -0,0 +1,28 @@
+import Component from '@ember/component';
+import { inject as service } from '@ember/service';
+import { tagName } from '@ember-decorators/component';
+import { action, computed } from '@ember/object';
+
+@tagName('')
+export default class SignInComponent extends Component {
+  @service
+  session
```
godfreyyeung: So the `oauth2-implicit-grant` authenticator will automatically fill out/update the session once it retrieves auth params?

allthesignals: Yup! It's a little bit too much "magic" but if you see the mixin in the application route, that mixin is responsible for communicating with the global "session" service.

PR #569, opened 2019-08-08
Make changes to integrate with new CRM-backed backend

julialhogan:
This PR is the bare minimum required changes to get ZAP search as it exists now working with the new frontend 

have tested locally running this version of UI against https://github.com/NYCPlanning/labs-zap-api/tree/zap-api-with-CRM and everything seems to work!!! 

deploying this for zap-search, and above branch as zap-api should result in everything that exists right now using CRM for projects details data source


allthesignals:
We need the CRM credentials for the dokku environment

PR #579, opened 2019-08-14
542: add lupp hearing form

### LINKED ISSUE #542: Build form for submitting hearing date details
[ screenshots tbd ]

Hearings are only required for project actions in this list: ZS, ZR, ZM, RS, PS, PQ, PP, PE, PC, NP, MM, ML, MC, BD, PX, HA, HC, HD, HO, HP, HU, BF, GF, PE, PI, ZA, ZC, SD, SC, SA, RA, RC, ME, MD, MY, CM, EC, HN, HG, HC 

- CRM form screenshots are included in #548. We can start with replicating these until the design is finalized.
- Hearing has one-to-one relationship with project
- build template for hearing form

trbmcginnis:
This PR creates a hearing form for lupp.

Users can enter:
- **date** --> using a calendar addon they can select any date in present or future
- **time** --> using ember-power-select and moment.js, user can select any time between 6AM and 11PM in intervals of 5 minutes
- **location** --> geocoded location using geosearch
- **additional location info** --> this is in case the geosearch autofill does not give the user a specific enough location, e.g. "the gym north of P.S. 8 main building" _Definitely can take suggestions on whether this is necessary/desired, or if someone has a suggestion for a better way to do this_

When users enter this information on the form, it is saved to Ember store (which is reset every time a user exits the route). When a user clicks on the "submit" button, everything is saved to the model through creating a new hearing record. 

**Starting Point for PR Review**: I would suggest starting at `/my-projects/project/hearing/add` for the route, template, and controller. 

Note: I tried to turn `autocomplete` off on each of the input fields, but geosearch is giving me issues on my chrome browser because google likes to autofill addresses--tried to investigate ways to turn it off but nothing seems to be working. Maybe we need to do this directly on `labs-ember-search`? Would love to hear thoughts

Closes #542


app/controllers/my-projects/project/hearing/add.js
```diff
@@ -0,0 +1,80 @@
...
+    this.onClearSearchResult();
+  }
+
+  // action triggered when click on "submit hearing" button
+  @action
+  saveHearingForm() {
+    // arguments to trigger a notification to the user that they need to input a field
+    if (!this.date) { this.set('dateMissing', true); } else { this.set('dateMissing', false); }
+    if (!this.location) { this.set('locationMissing', true); } else { this.set('locationMissing', false); }
+    if (!this.time) { this.set('timeMissing', true); } else { this.set('timeMissing', false); }
```
allthesignals: Another way this can be done is `this.set('dateMissing', !!this.date);`

app/controllers/my-projects/project/hearing/add.js
```diff
@@ -0,0 +1,80 @@
...
+export default class MyProjectsProjectHearingAddController extends Controller {
+  @service
+  store;
+
+  // for checking if user clicks submit button without inputting necessary attributes
+  dateMissing;
+
+  timeMissing;
+
+  locationMissing;
```
allthesignals: I'd set defaults for these, or make them computed from the model directly

app/controllers/my-projects/project/hearing/add.js
```diff
@@ -0,0 +1,80 @@
...
+      const locationLabel = this.location.label;
+      locationUpdated = locationLabel.concat('; ', this.additionalLocationInfo);
+    } else if (this.location) {
+      locationUpdated = this.location.label;
+    }
+
+    // if user has input date, location, and time then save to model and transition to to-review route
+    if (this.date && this.location && this.time) {
+      createHearing(this, locationUpdated, dateUpdated);
+      this.transitionToRoute('/my-projects/to-review');
```
allthesignals: So, if the `.save()` method fails (the server rejects the model), it'll still transition to the route — so the user might lose their work. You might make this an async await function, wrap `await createHearing(...)` in a `try catch` block, and catch on error.

app/controllers/my-projects/project/hearing/add.js
```diff
@@ -0,0 +1,80 @@
...
+    }
+  }
+}
```
allthesignals: This is great! One general comment I have about this: I would actually call `createRecord` on init, and then _directly pass the relevant model attributes where needed_... but since there are a lot of munging behaviors that need to happen (parsing dates, extract information from the location field), I would probably wrap all of this into a separate component and create actions for each field. My main concern is when someone might want to go back and edit a hearing for some reason — we wouldn't be able to re-use this.

app/router.js
```diff
@@ -35,6 +35,14 @@ Router.map(function() { // eslint-disable-line
     this.route('archive');
     this.route('upcoming');
     this.route('reviewed');
+
+    this.route('project', {
+      path: ':project_id',
+    }, function() {
```
allthesignals: sorry, tiny tiny nit, but this nested route syntax should look like the other nesting going on

app/routes/my-projects/project/hearing/add.js
```diff
@@ -0,0 +1,17 @@
...
+      controller.set('location', '');
+      controller.set('date', null);
+      controller.set('time', null);
+      controller.set('additionalLocationInfo');
+      controller.set('timeMissing', null);
+      controller.set('locationMissing', null);
+      controller.set('dateMissing', null);
+    }
+  }
+}
```
allthesignals: What happens if saving the model fails and the API rejects it? Then all that information will be lost.

app/routes/my-projects/to-review.js
```diff
@@ -1,4 +1,8 @@
 import Route from '@ember/routing/route';
 
 export default class MyProjectsToReviewRoute extends Route {
+  async model() {
+    const projects = this.store.findAll('project');
+    return projects;
```
allthesignals: tiny nit but you can just say `return this.store.findAll('project');` for brevity.

app/templates/my-projects/project/hearing/add.hbs
```diff
@@ -0,0 +1,70 @@
+<form>
+  <h3>{{model.dcp_projectname}}</h3>
+  <div class="grid-x">
+    <div class="large-12 cell">
+      <h5 class="no=margin">Hearing Date</h5>
+      {{#if dateMissing}}<label class="fix-hearing-info">*Please enter a valid date for your hearing</label>{{/if}}
```
allthesignals: I'd try to avoid in-lining this stuff

tests/unit/controllers/my-projects/project/hearing/add-test.js
```diff
@@ -0,0 +1,146 @@
...
+    };
+
+    // test that location attribute and date attribute in hearing model are empty at first
+    assert.equal(controller.model.hearing.get('location'), '');
+    assert.equal(controller.model.hearing.get('date'), null);
+
+    const hearingDate = new Date('2020-10-21T00:00:00'); // Wednesday, October 21, 2020
+    const hearingTime = new Date('1899-12-31T14:09:00'); // 2:09pm
+
+    // CASE 1: controller.location is empty
```
allthesignals: These different cases should be in separate test blocks `test('handle controller.location is empty')` — the reason being that you want to isolate each situation from the next. Duplication is totally fine tests.

trbmcginnis:
> Separately, I'd have added a full acceptance test for this, but the unit test covers the behavior well too. The problem with the unit test approach, however, is that it depends heavily on the implementation details of the feature. So, if we should need to change the implementation of the feature (refactor), we'd have to also update the test to match how we've implemented. So the advantage of an acceptance test is that you can refactor as much as you want but not have to update the test.

Ahh gotcha! I can def add an acceptance test :)

app/controllers/my-projects/project/hearing/add.js
trbmcginnis: cool cool, will edit

app/controllers/my-projects/project/hearing/add.js
trbmcginnis: Okay sounds good

app/controllers/my-projects/project/hearing/add.js
trbmcginnis: okay true

app/controllers/my-projects/project/hearing/add.js
trbmcginnis: hmm okay good point, I'll look into doing that

app/router.js
trbmcginnis: gotcha

app/routes/my-projects/project/hearing/add.js
trbmcginnis: Would this relate back to the comment about adding a try catch block?

app/routes/my-projects/to-review.js
trbmcginnis: yep

app/templates/my-projects/project/hearing/add.hbs
trbmcginnis: yep

tests/unit/controllers/my-projects/project/hearing/add-test.js
trbmcginnis: okay good to know! Will separate them out

godfreyyeung:
merged into `develop-future` manually

PR #626, opened 2019-09-04
595 & 596: Milestone DataStore Models, Mirage Models, factories, scenarios, updated templates

godfreyyeung:
One big change to note is that milestones are no longer side-loaded with projects as POJOs. 
In `routes/show-projects.js`, they are side-loaded as Ember Data Model instances..

### DIFF
--- a/app/routes/show-project.js
+++ b/app/routes/show-project.js
@@ -5,7 +5,7 @@ export default class ShowProjectRoute extends Route {
   async model({ id }) {
     const project = await this.store.findRecord('project', id, {
       reload: true,
-      include: 'actions',
+      include: 'actions,milestones',
     });
     return project;
   }
--- a/tests/integration/components/participant-types-test.js
+++ b/tests/integration/components/participant-types-test.js
@@ -3,7 +3,6 @@ import { setupRenderingTest } from 'ember-qunit';
 import { render } from '@ember/test-helpers';
 import hbs from 'htmlbars-inline-precompile';
 import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
-import seedMirage from '../../../mirage/scenarios/default';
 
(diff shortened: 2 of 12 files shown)


tests/integration/components/participant-types-test.js
```diff
@@ -3,24 +3,29 @@ import { setupRenderingTest } from 'ember-qunit';
...
     this.store = this.owner.lookup('service:store');
-    seedMirage(server);
   });
 
   test('it generates a list of participantTypes', async function(assert) {
-    // TODO: Retrieve by email?
-    this.user = await this.store.findRecord('user', 2, {
-      include: 'userProjectParticipantTypes.project,projects',
+    this.user = await this.store.createRecord('user');
+    this.project = await this.store.createRecord('project', { user: this.user });
```
allthesignals: ~Why store this on the function scope? If I'm not mistaken, the function scope is shared across other tests, which could cause memory issues.~ nm

app/routes/show-project.js
```diff
@@ -5,7 +5,7 @@ export default class ShowProjectRoute extends Route {
   async model({ id }) {
     const project = await this.store.findRecord('project', id, {
       reload: true,
-      include: 'actions',
+      include: 'actions,milestones',
```
allthesignals: Great!

tests/integration/components/participant-types-test.js
godfreyyeung: > @godfreyyeung this is looking great! One testing blindspot for us is the `show-project` page. I'll work on adding coverage there, but make sure you manually check that it still works in the development environment `ember s`. Thanks.

Yup, did a visual check and it looks okay! Ahh, thanks for helping with the test coverage on that page! 🙏

PR #628, opened 2019-09-05
Models lup

allthesignals:


app/templates/components/consolidated-cds.hbs
```diff
@@ -7,6 +7,7 @@
       {{#each boro.cds as |cd i|}}
         {{~if i ","}}
         {{#if cdlink}}
+          {{log boro.boroName}}
```
godfreyyeung: Can we delete this? I deleted it in a future commit, but I guess we could slip the deletion here, too

PR #633, opened 2019-09-09
631 & 632  - Hide dev features in staging and prod, hide rec+hearing btns if not auth'd

godfreyyeung:
- (#631) Adds ability to hide "Development" features when building to staging and production.
- (#632) This PR includes 1 rider: hiding rec+hearing buttons on `show-project` page if user not auth'd

### DIFF
--- a/app/templates/application.hbs
+++ b/app/templates/application.hbs
@@ -30,7 +30,9 @@
   {{#banner.nav}}
     <ul class="menu vertical large-horizontal">
       <li>{{link-to 'Search' 'show-geography'}}</li>
-      {{sign-in}}
+      {{#if (in-array (array "development" "test") this.ENV.environment)}}
+        {{sign-in}}
+      {{/if}}
     </ul>
   {{/banner.nav}}
 {{/labs-ui/site-header}}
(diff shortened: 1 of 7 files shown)


app/templates/application.hbs
```diff
@@ -30,7 +30,9 @@
   {{#banner.nav}}
     <ul class="menu vertical large-horizontal">
       <li>{{link-to 'Search' 'show-geography'}}</li>
-      {{sign-in}}
+      {{#if (in-array (array "development" "test") this.ENV.environment)}}
```
allthesignals: @trbmcginnis @godfreyyeung  Just FYI this will never show up in any production environment, so if we wanted to deploy this to a separate staging environment (ember s -e production) it wouldn't show any of these features. You could force a development deploy but that is weird.

What I meant by a "feature flag" was that the environment.js, which you have (`this.ENV.environment`), would look for an environmental variable (`ENABLE_LUPP_FEATURES=true ember s`).

At this point it doesn't quite matter but it will when we actually need to deploy these features to a staging environemnt....

godfreyyeung: Ohh I see. I was thinking we would just remove these conditionals when the sign-in is no longer "Development" ... but yeah this is not as good as the feature flag solution you mentioned.

allthesignals: @godfreyyeung no, don't close, please fix lint things and merge

allthesignals: @godfreyyeung This is fine but not what I had in mind. We need to move on so plz fix the lint things and merge. thanks

PR #635, opened 2019-09-09
Align actions for show project

allthesignals:
Forces the template to look for action models

### DIFF
--- a/app/models/action.js
+++ b/app/models/action.js
@@ -19,11 +19,11 @@ export default class ActionModel extends Model {
 
   // Name of action e.g. "Zoning Text Amendment"
   // sourced from dcp_name --> SUBSTRING(a.dcp_name FROM '-{1}\s*(.*)')
-  @attr('string') actionName;
+  @attr('string') name;
 
   // Action Code e.g. "ZR"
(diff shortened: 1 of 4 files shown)


app/models/action.js
```diff
@@ -19,11 +19,11 @@ export default class ActionModel extends Model {
 
   // Name of action e.g. "Zoning Text Amendment"
   // sourced from dcp_name --> SUBSTRING(a.dcp_name FROM '-{1}\s*(.*)')
-  @attr('string') actionName;
+  @attr('string') name;
 
   // Action Code e.g. "ZR"
   // sourced from dcp_name --> SUBSTRING(a.dcp_name FROM '^(\w+)')
```
godfreyyeung: we should probably add comment clarifying it maps to ZAP-API:actioncode 🙈

app/models/project.js
```diff
@@ -77,7 +75,9 @@ export default class ProjectModel extends Model {
   @attr() bbls;
 
   @attr({ defaultValue: () => EmptyFeatureCollection })
-  bbl_featurecollection;
+  bbl_featurecollection
+
+  @attr() milestones;
```
godfreyyeung: Why going back to POJO milestones? 
oh --- I'm guessing we're leaving that to a later commit?

PR #637, opened 2019-09-09
625: put hearing form into own component, update directly on model

### LINKED ISSUE #625: Separate out Hearing Form into component and revise models
- Create own component for hearing form instead of just using route
- Assure that changes to hearing attributes are mutating model directly

trbmcginnis:
This PR:

- puts hearing form into own component
- updates how hearing form interacts with model

Next steps:

- updating the hearing models relationship with other models
- updating the time component to reflect a more user-friendly layout

Closes #625


app/components/hearing-form.js
```diff
@@ -0,0 +1,85 @@
+import Component from '@ember/component';
+import { action, computed } from '@ember/object';
+import moment from 'moment';
+
+export default Component.extend({
```
allthesignals: we should use class syntax:

`export default class HearingFormComponent extends Component { }`

app/components/hearing-form.js
```diff
@@ -0,0 +1,85 @@
...
+  // to check whether hearing model attributes "location" and "date" are populated
+  hearingSubmitted: null,
+
+  // minimum date for calendar
+  currentDate: new Date(),
+
+  // for clearing search results on the geosearch input
+  onClearSearchResult: () => {},
+
+  @computed()
```
allthesignals: I'm surprised this syntax still works — decorators for object literal properties were part of an old proposal. Anyways, it's deprecated now, and we should be using ES6 classes (see above) when using decorators.

app/components/hearing-form.js
allthesignals: it was es6 class syntax before when it was a controller — why use object literal syntax in the component?

allthesignals:
Plz fix tests @trbmcginnis

PR #638, opened 2019-09-09
Fix milestone attr names to align with CRM

godfreyyeung:
Part of 571...
This PR's commit is almost the same as `571-alignment-milestones`. The small differences include:
  - change `projectMilestone` attr to `milestone`
  - Retain camelcasing for the `milestoneLinks` attribute, since zap-api provides the camelcasing.

### DIFF
--- a/app/models/project.js
+++ b/app/models/project.js
@@ -26,7 +26,7 @@ export default class ProjectModel extends Model {
   // ONE Project Has Many User Project Participant Types
   @hasMany('userProjectParticipantType') userProjectParticipantTypes;
 
-  @hasMany('miletone') milestones;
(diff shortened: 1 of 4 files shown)


app/models/project.js
```diff
@@ -26,6 +26,8 @@ export default class ProjectModel extends Model {
   // ONE Project Has Many User Project Participant Types
   @hasMany('userProjectParticipantType') userProjectParticipantTypes;
 
+  @hasMany('miletone') milestones;
```
allthesignals:
```suggestion
  @hasMany('milestone', { async: false }) milestones;
```

PR #643, opened 2019-09-10
Use explicit LUPP_ENABLED flag for deployments

allthesignals:
Adds an explicit LUPP_ENABLED flag for disabling/enabling LUPP features within ZAP.

Removes an unnecessary serializer. Removes made up environments, which we were relying on to switch host information.

Because "test", "development", and "production" and the only meaningful environment names in Ember (meaningful in that they tell Ember how to build the app), we should be using environment variables for specific things like the host API.

So, **if you need to have the app consume a different API, you must say `HOST_API=http://something:something ember s`**.

The Netlify configuration has been updated accordingly.


mirage/config.js
```diff
@@ -39,10 +40,16 @@ export default function() {
     const { id } = request.params;
     const { me } = request.queryParams;
 
+    if (id) {
+      return schema.users.find(id); // users in the second case
+    }
+
     // check that this is a request for an auth'd user and there's a cookie
-    if (me && document.cookie.includes('token')) return schema.users.first();
+    if (me) {
```
godfreyyeung: No more check for a cookie here? Why not? 
Should we also update the comment?

allthesignals: @godfreyyeung Yep!

allthesignals:
> > Because "test", "development", and "production" and the only meaningful environment names in Ember (meaningful in that they tell Ember how to build the app), we should be using environment variables for specific things like the host API.
> 
> Do you mean that we shouldn't use the environment variable, which is either "test" "development" or "production", because it's too much of a "blanket solution"? And a feature flag would be more specific to just toggling features

Yeah, more of less, but this concerns how the host name API for the adapters is specified. For example, there are cases where we want our app to point to staging API, but have it build as a production Ember app (as in it will minify, transpile, etc). Before, we were specifying environments other than the 3, and I think in that case it defaults to development.

mirage/config.js
allthesignals: @godfreyyeung because we're using http-only cookies, and because you can't simulate http-only cookies in mirage (because mirage is executed in the browser, which cannot touch http-only cookies), we just have to assume that the API handles the cookies correctly.

Issue #650, opened 2019-09-11
Redesign Hearing page to allow adding hearings by action


godfreyyeung:
See Andy's wireframes, ask Andy or Godfrey for clarification. 

 - User can either select (1) "Submit single Hearing for all actions" or (2) "Submit Hearing for each action"
   - If (2) and in real life one hearing applies to multiple (but not all) actions, then the user must manually copy and paste that hearing's info into the inputs under each of those actions. 

- If (2),  User MUST fill out all hearing values for actions 
(at least, in this MVP iteration)

PR #669, opened 2019-09-18
650: revise hearing form to match new disposition model

trbmcginnis:
- create `date-time-picker` and `location-input` components
- refactor `hearing-form` to match the new disposition model
- provide users with option to submit one hearing for ALL actions in a project or one hearing PER action

Addresses issues #650


app/components/location-input.js
```diff
@@ -0,0 +1,21 @@
...
+  @action
+  setLocation() {
+    // for checking whether the user has entered text for a location
+    // for displaying invalid inputs, the argument checkIfMissing is passed down
+    // from the template for the route, add.hbs
+    // checkIfMissing is set to true every time the "submitHearing" runs
+    // e.g. {{#if (and locationMissing checkIfMissing)}}
+    this.set('locationMissing', !this.location);
+
+    this.set('targetField', this.location);
```
allthesignals: I would add a class prop called targetField

app/routes/my-projects/project/hearing/add.js
```diff
@@ -1,4 +1,28 @@
...
+  // Depends on the my-project/project route already loading a project
+  // with side-loaded actions
+
+  @service
+  store;
+
+  @service
+  currentUser;
+
+  model() {
```
allthesignals: `async model() {`

app/routes/my-projects/project/hearing/add.js
```diff
@@ -1,4 +1,28 @@
...
+    return this.modelFor('my-projects.project');
+  }
+
+  async setupController(controller, model) {
+    super.setupController(controller, model);
+    controller.set('dispositions', model.dispositions);
+  }
```
allthesignals: We might wanna nix this if it's not being used

app/controllers/my-projects/project/hearing/add.js
```diff
@@ -1,4 +1,106 @@
...
+  @action
+  async submitHearing() {
+    const model = this.get('model');
+    const allActions = this.get('allActions');
+
+    try {
+      if (allActions) {
+        const projectDispositions = model.dispositions;
+        const allDispositionsLocation = this.get('dispositionForAllActions.publichearinglocation');
+        const allDispositionsDate = this.get('dispositionForAllActions.dateofpublichearing');
```
allthesignals: Could we use object destructing?

app/controllers/my-projects/project/hearing/add.js
```diff
@@ -1,4 +1,106 @@
...
+    const allActions = this.get('allActions');
+
+    // a function to check if each hearing location/date field is truthy
+    function infoExists(hearingInfo) {
+      return hearingInfo;
+    }
+
+    if (allActions) {
+      const dispositionLocation = dispositionForAllActions.publichearinglocation;
+      const dispositionDate = dispositionForAllActions.dateofpublichearing;
```
allthesignals: Object destructuring

app/controllers/my-projects/project/hearing/add.js
```diff
@@ -1,4 +1,106 @@
...
+
+  modalOpen = false;
+
+  dispositionForAllActions = DispositionForAllActions.create();
+
+  // passed down to date-time-picker and location-input components
+  // set to true when the user clicks on Submit Hearing button
+  // and "submitHearing" action sets it to true
+  // checkIfMissing is for notifying the user which inputs are invalid
+  checkIfMissing = null;
```
allthesignals: Should this be `false` by default?

allthesignals: nvm - we want them to be not filled in at first

app/controllers/my-projects/project/hearing/add.js
```diff
@@ -1,4 +1,106 @@
...
+
+  @action
+  closeModal() {
+    this.set('modalOpen', false);
+  }
+
+  // checks that the user had input valid values
+  // opens up the modal so the user can review their inputs
+  @action
+  checkInfo() {
```
allthesignals: Can we change the naming of this method? Like "onSubmit"?

app/controllers/my-projects/project/hearing/add.js
```diff
@@ -1,4 +1,106 @@
...
+        const allDispositionsDate = this.get('dispositionForAllActions.dateofpublichearing');
+
+        projectDispositions.forEach(function(disposition) {
+          disposition.set('publichearinglocation', allDispositionsLocation);
+          disposition.set('dateofpublichearing', allDispositionsDate);
+        });
+
+        await model.dispositions.save();
+      } else {
+        await model.dispositions.save();
```
allthesignals: Let's try to simulate a network failure and see what happens _ I think it will probably go ahead and say "hearingSubmitted" true :) and we don't want that

allthesignals: Or actually that might have to do with using `fetch` directly

app/controllers/my-projects/project/hearing/add.js
```diff
@@ -1,4 +1,106 @@
...
+      }
+      // for displaying certain elements
+      this.set('hearingSubmitted', true);
+      this.set('modalOpen', false);
+    } catch (e) {
+      alert('Sorry about that, unable to connect to server, can you try again?'); // eslint-disable-line
```
godfreyyeung: I'm also averse to using `alert()`, since, like modals, serializes the user's UI.  
In recommendations I assign the error message to a property on the controller. E.g.
```
// ../add.js
this.set('error', 'Sorry about that, unable to connect to server, can you try again?')
```
And then in the template, I render this error if its set
```
// ../add.hbs
{{#if this.error}}
    {{this.error}}
{{/if}}
```
@allthesignals  thoughts?

allthesignals: @godfreyyeung that works — you might even just set the error object itself. We don't know if it's a connection issue or an authentication. See other parts of the source code that extract error messages from a `catch`

allthesignals:
@trbmcginnis this is great! Let's try to get some things addressed (move some stuff into computeds, etc). If it takes too long let's just move on but create new issues

tests/acceptance/user-can-fill-out-hearing-form-test.js
```diff
@@ -4,68 +4,303 @@ import {
...
 
-    // because the user has not input a date/time/location yet, the submit button will still appear
-    assert.ok(find('[data-test-button="hearingSubmit"]'));
+    // user fills in a location input
+    await fillIn('[data-test-allactions-input="location"]', '121 Bananas Ave, Queens, NY');
+
+    // user triggers a keyup event (necessary for acceptance test)
+    triggerEvent('[data-test-allactions-input="location"]', 'keyup', {
+      keyCode: 84, // t
+    });
```
allthesignals: I think this needs to be awaited

app/controllers/my-projects/project/hearing/add.js
andycochran: Is `onSubmit` right tho? This action doesn't actually submit, it just validates and opens the modal. Submission happens via another action on a button in the modal.

trbmcginnis:
Made all the changes except for the error handling, going to address that in a separate issue. Going to go ahead and merge.

Issue #619, opened 2019-08-26
Add recommendation submission confirmation screen


hannahkates:

andycochran:
I think this is done.

PR #677, opened 2019-09-18
648 Disposition based recommendation submission

godfreyyeung:
TODO for upcoming PRs: 
- [ ] Thorough tests
- [ ] Validation
- [ ] pre-submission confirmation dialogue #619 #674 
- [ ] post-submission navigation #680 
- [ ] Switch to using `disposition.participantType` instead of queryParams

### DIFF
--- a/app/controllers/my-projects/project/recommendations/add.js
+++ b/app/controllers/my-projects/project/recommendations/add.js
@@ -1,15 +1,35 @@
 import Controller from '@ember/controller';
-import EmberObject, { action, computed } from '@ember/object';
+import EmberObject, { action } from '@ember/object';
 import { alias } from '@ember/object/computed';
 import { inject as service } from '@ember/service';
-import { A } from '@ember/array';
 
-class RecommendationOption extends EmberObject {
-  label = '';
+const RECOMMENDATION_FIELD_BY_PARTICIPANT_TYPE_LOOKUP = {
+  BB: 'boroughboardrecommendation',
+  BP: 'boroughpresidentrecommendation',
+  CB: 'communityboardrecommendation',
+};
 
-  code = '';
+// all attributes in this class map to their exact match in the Disposition model
+class DispositionForAllActions extends EmberObject {
+  // the selected recommendation option if applying filled Recommendation
+  // to all actions.
+  // On submission, this is mapped to the correct Disposition model attribute. ie.
+  // communityboardrecommendation, boroughboardrecommendation or boroughpresidentrecommendation
+  recommendation = '';
 
-  actions = A();
+  votinginfavorrecommendation = 0;
+
+  votingagainstrecommendation = 0;
+
+  votingabstainingonrecommendation = 0;
+
+  totalmembersappointedtotheboard = 0;
+
+  votelocation = '';
+
+  dateofvote = '';
+
+  consideration = '';
 }
 
 export default class MyProjectsProjectRecommendationsAddController extends Controller {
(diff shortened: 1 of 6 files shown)


app/controllers/my-projects/project/recommendations/add.js
```diff
@@ -25,58 +39,80 @@ export default class MyProjectsProjectRecommendationsAddController extends Contr
...
+    const thisCtrl = this;
+    this.dispositions.forEach(function(disposition) {
+      if (thisCtrl.allActions) {
+        thisCtrl.send('setDispositionRecByPartType', disposition, thisCtrl.dispositionForAllActions.recommendation);
+        disposition.set('votinginfavorrecommendation', thisCtrl.dispositionForAllActions.votinginfavorrecommendation);
+        disposition.set('votingagainstrecommendation', thisCtrl.dispositionForAllActions.votingagainstrecommendation);
+        disposition.set('votingabstainingonrecommendation', thisCtrl.dispositionForAllActions.votingabstainingonrecommendation);
+        disposition.set('totalmembersappointedtotheboard', thisCtrl.dispositionForAllActions.totalmembersappointedtotheboard);
+        disposition.set('consideration', thisCtrl.dispositionForAllActions.consideration);
+      }
```
allthesignals: You can do this with setProperties and it'd be much more legible. If you're concerned about using POJOs in the future, you can import setProperties from `@ember/object` and use it like this:
```
setProperties(disposition, {
  // ...
});
```

app/controllers/my-projects/project/recommendations/add.js
```diff
@@ -25,58 +39,80 @@ export default class MyProjectsProjectRecommendationsAddController extends Contr
...
+ * @param { Disposition } disposition
+ * @param { String } recommendation
+ * @param { String } participantType (optional)
+ * assigns the `recommendation` string to either the `
+ * TODO: Update this to rely on the disposition.participantType field
+ * when it is implemented in the ZAP-API.
+ */
+  @action
+  setDispositionRecByPartType (disposition, recommendation) {
+    const partType = this.participantType;
```
allthesignals: Just use object destructing so it's easier to follow:

`const { participantType } = this;`

No need to create new names for things...

app/controllers/my-projects/project/recommendations/add.js
```diff
@@ -25,58 +39,80 @@ export default class MyProjectsProjectRecommendationsAddController extends Contr
...
+    let recommendationFieldName = '';
+    if (partType === 'CB') {
+      recommendationFieldName = 'communityboardrecommendation';
+    } else if (partType === 'BB') {
+      recommendationFieldName = 'boroughboardrecommendation';
+    } else if (partType === 'BP') {
+      recommendationFieldName = 'boroughpresidentrecommendation';
+    } else {
```
allthesignals: This is a lookup. Let's move this lookup to a constant above this controller class:

```
const PARTICIPANT_TYPE_FIELD_LOOKUP = {
  BB: 'boroughboardrecommendation',
  BP: 'boroughpresidentrecommendation',
};
```

allthesignals: Then we can just say
```
const targetField = PARTICIPANT_TYPE_FIELD_LOOKUP[participantType];
disposition.set(targetField, recommendation);
```

app/templates/my-projects/project/recommendations/add.hbs
```diff
@@ -10,61 +10,372 @@
...
+                    @value={{this.dispositionForAllActions.totalmembersappointedtotheboard}}
+                    @key-up={{fn (mut this.dispositionForAllActions.totalmembersappointedtotheboard) value="target.value"}}
+                  />
+                </div>
+              </div>
+            {{/if}}
+          </fieldset>
+        </div>
+      {{/if}}
+      {{#if (eq this.allActions false)}}
```
allthesignals: Wait... `allActions` is a boolean, no? So this could just be `{{#if this.allActions}}`, right?

app/templates/my-projects/project/recommendations/add.hbs
```diff
@@ -10,61 +10,372 @@
...
             />
-            <label for="quorum-no">
-              No
+            <label for="all-actions-no">
+              No, I will submit a recommendation for each action
             </label>
           </fieldset>
+        </div>
+      {{/if}}
+      {{#if (eq this.allActions true)}}
```
allthesignals: `{{#if this.allActions}}`

app/templates/my-projects/project/recommendations/add.hbs
godfreyyeung: I wanted to make it a strict equals to `false`, because `this.allActions` starts out as `undefined`. 
I wanted to only show the "By Action" Recommendation forms after the user has selected either Yes or No.

andycochran: It's `null` by default, right? So we explicitly want to know if it's `true` or `false`.

app/controllers/my-projects/project/recommendations/add.js
godfreyyeung: ✅

app/controllers/my-projects/project/recommendations/add.js
godfreyyeung: ✅

app/controllers/my-projects/project/recommendations/add.js
godfreyyeung: ✅

PR #684, opened 2019-09-20
Disallow access to the my-projects route if not logged in

### LINKED ISSUE #683: Disallow access to protected routes if unauthorized

allthesignals:
Closes #683

### DIFF
--- a/app/routes/my-projects.js
+++ b/app/routes/my-projects.js
@@ -1,4 +1,7 @@
 import Route from '@ember/routing/route';
+import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
 
-export default class MyProjectsRoute extends Route {
+const AuthenticatedRoute = Route.extend(AuthenticatedRouteMixin);
+
+export default class MyProjectsRoute extends AuthenticatedRoute {
 }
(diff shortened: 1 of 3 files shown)


app/controllers/show-geography.js
```diff
@@ -159,6 +159,8 @@ export default class ShowGeographyController extends GeographyParachuteControlle
       this.transitionToRoute('oops');
     }
 
+    console.log(projects);
```
godfreyyeung: oops, should remove

app/routes/my-projects.js
```diff
@@ -1,4 +1,7 @@
 import Route from '@ember/routing/route';
+import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
 
-export default class MyProjectsRoute extends Route {
+const AuthenticatedRoute = Route.extend(AuthenticatedRouteMixin);
```
godfreyyeung: will we also need to add this to app/routes/show-project.js?

app/controllers/show-geography.js
allthesignals: thx

app/routes/my-projects.js
allthesignals: @godfreyyeung nope — you don't need to be authenticated to look at the whole route. You only need to be authenticated to see the buttons for hearings/rec. Right?

PR #692, opened 2019-09-20
improve the layout and markup of the recommendation form

andycochran:


app/templates/my-projects/project/recommendations/add.hbs
```diff
@@ -39,471 +39,434 @@
             <p>{{project.dcp_projectbrief}}</p>
           </div>
         </div>
-      </div>
```
allthesignals: There was an issue visually with the # of breaks beteween actions but moving on

PR #713, opened 2019-09-24
662: use real lup review start and end dates

hannahkates:
This PR is working, but I'm having trouble making 100% sure the calculations appear as preferred since the mirage data doesn't provide realistic start and end dates for the review milestone (and the API and front-end aren't properly connected to show real projects in "to review").


godfreyyeung:
Replacing this PR with https://github.com/NYCPlanning/labs-zap-search/pull/734

PR #719, opened 2019-09-26
718: Put confirmation modal into own component

trbmcginnis:
Modal for prompting user to review and then confirm or cancel their hearing & rec inputs

### DIFF
--- a/app/templates/components/confirmation-modal.hbs
+++ b/app/templates/components/confirmation-modal.hbs
@@ -1,0 +1,7 @@
+{{#if open}}
+  <div class="reveal-overlay" style="display:block;">
+    <div class="reveal" role="dialog" aria-hidden="false" tabindex="-1" style="display:block;">
+        {{yield}}
+    </div>
+  </div>
+{{/if}}
--- a/app/templates/my-projects/project/recommendations/add.hbs
+++ b/app/templates/my-projects/project/recommendations/add.hbs
@@ -463,109 +463,105 @@
               </label>
               <input type="file" id="exampleFileUpload_2" class="show-for-sr">
             </div>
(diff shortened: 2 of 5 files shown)


godfreyyeung:
Is it possible to move the `modalOpen` variable inside the `confirmation-modal` component? 
The modal component can then keep  track of `modalOpen` by itself.

To let outside controllers/templates interact with the modal, the modal could yield some actions for toggling `modalOpen`

@allthesignals thoughts on component design?

app/templates/components/confirmation-modal.hbs
```diff
@@ -0,0 +1,5 @@
+<div id="hearing-confirmation" class="reveal-overlay" style="display:block;">
```
andycochran: The `id` should be more generic if this is going to be used in places other then the hearing form.

andycochran: Actually, IDK if an `id` is even necessary.

app/templates/my-projects/project/recommendations/add.hbs
```diff
@@ -464,107 +464,105 @@
...
                   {{/if}}
-                  {{#if this.allActions}}
-                    <h5>
-                      Recommendation:
+                {{else}}
+                  {{#each this.dispositionAndChangesetPairs as |dispositionChangesetPair idx|}}
+                    <h5 class="large-margin-top">
+                      Recommendation: {{dispositionChangesetPair.disposition.action.name}}
+                      <small class="dark-gray">
+                        ({{dispositionChangesetPair.disposition.action.ulurpnumber}})
```
allthesignals: ulurpnumber <- needs to be dcp

Issue #662, opened 2019-09-16
Calculate days remaining in LUP's review milestone


hannahkates:
Days remaining = actual end date - today's date

allthesignals:
This says there was something merged, but when is it finished? Should this go into staging? @hannahkates

hannahkates:
@allthesignals The merged PR was for the related API work. Front-end PR is forthcoming.

PR #727, opened 2019-09-30
709 New Project attributes and real CB milestone mirage data

godfreyyeung:
1) Introduces new attributes to the Project model to support displaying milestone-based dates on project cards.
2) Adds Mirage data to support those new attributes. 
3) However the data is only fleshed out for the CB user. 

This is a step towards @andycochran 's wireframes for the tabs

This PR covers some of #662 

Riders:
- lots of misc. fixes to milestone property names, planned/actual date values, milestone ids etc

### DIFF
--- a/app/models/project.js
+++ b/app/models/project.js
@@ -16,6 +16,16 @@ const EmptyFeatureCollection = {
   }],
 };
 
+const COMMUNITY_BOARD_REFERRAL_MILESTONE_ID = '923beec4-dad0-e711-8116-1458d04e2fb8';
+const BOROUGH_PRESIDENT_REFERRAL_MILESTONE_ID = '943beec4-dad0-e711-8116-1458d04e2fb8';
+const BOROUGH_BOARD_REFERRAL_MILESTONE_ID = '963beec4-dad0-e711-8116-1458d04e2fb8';
+
+const MILESTONE_ID_LOOKUP = {
+  CB: COMMUNITY_BOARD_REFERRAL_MILESTONE_ID,
+  BP: BOROUGH_PRESIDENT_REFERRAL_MILESTONE_ID,
+  BB: BOROUGH_BOARD_REFERRAL_MILESTONE_ID,
+};
+
 export default class ProjectModel extends Model {
   // Many Users to Many Projects
   @hasMany('user') users;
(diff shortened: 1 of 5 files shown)


app/models/project.js
```diff
@@ -16,6 +16,16 @@ const EmptyFeatureCollection = {
...
+const COMMUNITY_BOARD_REFERRAL_MILESTONE_ID = '923beec4-dad0-e711-8116-1458d04e2fb8';
+const BOROUGH_PRESIDENT_REFERRAL_MILESTONE_ID = '963beec4-dad0-e711-8116-1458d04e2fb8';
+const BOROUGH_BOARD_REFERRAL_MILESTONE_ID = '943beec4-dad0-e711-8116-1458d04e2fb8';
+
+const MILESTONE_ID_LOOKUP = {
+  CB: COMMUNITY_BOARD_REFERRAL_MILESTONE_ID,
+  BP: BOROUGH_PRESIDENT_REFERRAL_MILESTONE_ID,
+  BB: BOROUGH_BOARD_REFERRAL_MILESTONE_ID,
+};
+
```
allthesignals: Great!

app/models/project.js
```diff
@@ -31,8 +41,80 @@ export default class ProjectModel extends Model {
...
+  @attr() teammemberrole;
+
   @attr() applicantteam;
 
+  @computed('milestones')
+  get publicReviewPlannedStartDate() {
+    const cbMilestone = this.milestones.find(milestone => milestone.milestone === COMMUNITY_BOARD_REFERRAL_MILESTONE_ID);
+    if (cbMilestone) {
+      return cbMilestone.plannedstartdate;
```
allthesignals: I would destructure this from line 53:

`const { plannedstartdate } = ...`

app/models/project.js
```diff
@@ -31,8 +41,80 @@ export default class ProjectModel extends Model {
...
+    }
+    return null;
+  }
```
allthesignals: Does this computed return a string? What happens when you pass this through moment.js or something that expects a string, and it's null?

app/models/project.js
```diff
@@ -31,8 +41,80 @@ export default class ProjectModel extends Model {
...
+
+  // These displayMilestoneActual<Start/End>
+  // date fields represent different dates from different milestones
+  // depending on `tab`.
+  // If `tab` is 'upcoming' or 'to-review', these dates are derived from
+  // the participant's (specified by `teammemberrole`) review milestone.
+  // If `tab` is 'reviewed'...
+  //   - these dates come from the current In Progress milestone
+  //   - if there is two In Progress milestones, an array of milestone dates is returned
```
allthesignals: I like this - thanks!

app/models/project.js
```diff
@@ -31,8 +41,80 @@ export default class ProjectModel extends Model {
...
+  @computed('tab', 'teammemberrole', 'milestones')
+  get displayMilestoneActualStartDate() {
+    if (this.tab === 'to-review') {
+      const participantMilestoneId = MILESTONE_ID_LOOKUP[this.teammemberrole];
+      const participantReviewMilestone = this.milestones.find(milestone => milestone.milestone === participantMilestoneId);
+      return participantReviewMilestone.actualstartdate;
```
allthesignals: prefer destructure from line 72

app/models/project.js
```diff
@@ -31,8 +41,80 @@ export default class ProjectModel extends Model {
...
+    }
+    if (this.tab === 'reviewed') {
+      const inProgressMilestones = this.milestones.filter(milestone => milestone.statuscode === 'In Progress');
+      return inProgressMilestones.map(milestone => ({
+        milestone: milestone.milestone,
+        actualstartdate: milestone.actualstartdate,
+      }));
+    }
+    return null;
+  }
```
allthesignals: My concern with this computed is that it can return 3 possible values: null, string, or array...

What if we made new computed for a "matrix" of possible outcomes? For example, computeds: `reviewedInProgressMilestones`, `toReviewStartDates`.

The tab property (where does that come from) seems like _view_ logic (although that state is derived from something like "LUP Status"), and therefore, doesn't belong in the model. Perhaps the template or something downstream (a component) could switch to the correct property. 

My reasoning behind this is that there's a lot of different possible outcomes for all of these computeds, you know? So something unexpected could happen.

Up to you whether you wanna change it!

app/models/project.js
```diff
@@ -31,8 +41,80 @@ export default class ProjectModel extends Model {
...
+    }
+    if (this.tab === 'reviewed') {
+      const inProgressMilestones = this.milestones.filter(milestone => milestone.statuscode === 'In Progress');
+      return inProgressMilestones.map(milestone => ({
+        milestone: milestone.milestone,
+        actualenddate: milestone.actualenddate,
+      }));
+    }
+    return null;
+  }
```
allthesignals: Yeah, similar to my comment above, we have a lot of these "display" type computed properties. Are they used only in a specific component? What do you think about this: distill from here the actual raw data that informs those views AND the component-specific data structures (for example, this `map` into an array of objects... is that a data structure that's expected by a component somewhere? These seem to know too much about each other...

app/models/project.js
```diff
@@ -31,8 +41,80 @@ export default class ProjectModel extends Model {
...
+      if (this.dcp_publicstatus_simp === 'filed') {
+        return this.publicReviewPlannedStartDate;
+      }
+      const participantMilestoneId = MILESTONE_ID_LOOKUP[this.teammemberrole];
+      const participantReviewMilestone = this.milestones.find(milestone => milestone.milestone === participantMilestoneId);
+      return participantReviewMilestone.plannedstartdate;
+    }
+    return null;
+  }
+
```
allthesignals: Ditto ^

mirage/scenarios/default.js
```diff
@@ -532,275 +699,161 @@ export default function(server) {
...
+  });
+
+  server.create('milestone', 'boroughPresidentReview', {
+    project: seedBPUserProjects[0],
+    statuscode: 'Not Started',
+    actualstartdate: moment().add(19, 'days'),
+    displayDate: moment().add(19, 'days'),
+    actualenddate: null,
+    displayDate2: null,
+  });
```
allthesignals: This scenario file is getting really long. Do you think this would be better addressed with Mirage fixtures? -> https://www.ember-cli-mirage.com/docs/data-layer/fixtures

app/models/project.js
allthesignals: Check this out: https://en.wikipedia.org/wiki/Cyclomatic_complexity

app/models/project.js
godfreyyeung: I'm designing it so that it's up to the template or controller to check whether the computed returns a string or is null. Thinking that returning a null value is useful for indicating when publicReviewPlannedStartDate (or any of the other date fields) is not applicable.

godfreyyeung: Although returning `null` is more appropriate for the `displayMilestoneActual<Start/End>Date` fields. Theoretically there should always be a community board review planned start date?

godfreyyeung: seems `{{moment-format null}}` yields "Invalid date"

app/models/project.js
godfreyyeung: Ohh, I like the idea of breaking out the computeds down even more. So then each one can return at most two values (e.g. null, array). 

The tab property was part of the "new project schema" from our brainstorming with Hannah a few weeks back. I think the goal was to help relieve the frontend from some filtering logic for the tabs. 

I like this suggestion and think it's worthwhile! I'm going to open a new issue for it and tackle it separately. 

I think after writing out PR/Issue #728 , which depends on this PR, I'll have even more clear hindsight on how to best approach this.

app/models/project.js
godfreyyeung: I like this idea a lot. Makes sense -- breaking the computed down into two tab-specific computeds, one for "to-review" and another for "reviewed", and tying each computed to a tab-specific component.

mirage/scenarios/default.js
godfreyyeung: yes! I had the same feeling, and was wondering if fixtures would help. I'll open an issue to look into it.

app/models/project.js
godfreyyeung: JK I think I can just include it in this PR 😂

PR #735, opened 2019-10-02
732 Handle single disposition on Recommendation form

godfreyyeung:
# What
Automatically presents the "one dispositions for all actions" recommendation form if there is only one disposition.

It also creates an extra mirage CB user project to test the scenario of a "to-review" project only having one dispo.

Thus I had to update some tests that depended on a certain sequence of projects/dispositions.

# To think about
The tight coupling between our tests and our one `scenarios/default.js` mirage file is becoming unwieldy. We should maybe start prioritizing setting up different scenarios and/or setting up test-suite-specific  mirage data 

Thoughts @allthesignals @trbmcginnis ?

### DIFF
--- a/app/routes/my-projects/project/recommendations/add.js
+++ b/app/routes/my-projects/project/recommendations/add.js
@@ -19,6 +19,7 @@ export default class MyProjectsProjectRecommendationsAddRoute extends Route {
   async setupController(controller, model) {
     super.setupController(controller, model);
     controller.set('dispositions', model.dispositions);
+    controller.set('allActions', model.dispositions.length <= 1 ? true : null);
   }
 
   @action
--- a/mirage/factories/project.js
+++ b/mirage/factories/project.js
@@ -177,10 +177,4 @@ export default Factory.extend({
       },
     ];
   },
-
-  afterCreate(project, server) {
-    // The number of actions created here cannot be more than the number of unique
-    // action codes in the action factory.
-    server.createList('action', 7, { project });
-  },
 });
--- a/mirage/scenarios/default.js
+++ b/mirage/scenarios/default.js
@@ -1,6 +1,6 @@
 import moment from 'moment';
 
-const NUM_CB_USER_PROJECTS = 7;
+const NUM_CB_USER_PROJECTS = 8;
 const NUM_BP_USER_PROJECTS = 4;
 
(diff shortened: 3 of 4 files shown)


allthesignals:
What's this? Issue #?

godfreyyeung:
> What's this? Issue #732 ?

yup yup

app/routes/my-projects/project/recommendations/add.js
```diff
@@ -19,6 +19,7 @@ export default class MyProjectsProjectRecommendationsAddRoute extends Route {
   async setupController(controller, model) {
     super.setupController(controller, model);
     controller.set('dispositions', model.dispositions);
+    controller.set('allActions', model.dispositions.length <= 1 ? true : null);
```
allthesignals: So, I think these should be computeds on the controller or a component — for example, what if the # of dispositions change? the value of `allActions` will be out of sync.

godfreyyeung: ohhh. Hmm. I don't think the number of dispositions would ever change within one user's browsing session though. I thought it's predetermined by the CRM.  I guess if somehow a planner added more actions via CRM while a user was using the website, it could happen between page loads... hmmm

allthesignals:
> The tight coupling between our tests and our one scenarios/default.js mirage file is becoming unwieldy. We should maybe start prioritizing setting up different scenarios and/or setting up test-suite-specific mirage data

@godfreyyeung absolutely — in fact we should almost always be coupling the data context with the thing under test. The issue we run into here is a testing antipattern called spooky action at a distance—things are happening or not happening in our scenario/default.js file that cause tests to fail

godfreyyeung:
> > The tight coupling between our tests and our one scenarios/default.js mirage file is becoming unwieldy. We should maybe start prioritizing setting up different scenarios and/or setting up test-suite-specific mirage data
> 
> @godfreyyeung absolutely — in fact we should almost always be coupling the data context with the thing under test. The issue we run into here is a testing antipattern called spooky action at a distance—things are happening or not happening in our scenario/default.js file that cause tests to fail

@allthesignals roger. At first I was testing if using the scenario file as a testing "bootstrap" would be passable, but now I clearly see it has way too much overhead.

Issue #627, opened 2019-09-04
Create and plug in mirage data for Archive tab


godfreyyeung:
- Use Mirage data for development

- [x] Upcoming
- [x] To Review
- [x] Reviewed
- [ ] Archive

allthesignals:
This seems like it's in a done state, no?

hannahkates:
The "Archive" tab still need to converted to using mirage data instead of hardcoded.

PR #737, opened 2019-10-02
627: plug in mirage data on archive tab

hannahkates:
This PR:
- Updates the "Archive" template and route so that it uses project and milestone model values
- Adds two new attributes to the project model. **Note: These will require API changes, since neither of them are currently provided by the API (Attn: @allthesignals )**
  - `publicStatus` maps to `dcp_publicstatus` in the CRM project entity. It provides values like 'Approved' or 'Withdrawn' which the API recodes into `dcpPublicstatusSimp` = 'Completed'. We need more granularity than the simplified 'Completed' value provided in `dcpPublicstatusSimp`
  - `projectCompleted` maps to `dcp_projectcompleted` in the CRM project entity. It provides the date that the project became complete
- Updates mirage data to provide attributes for `publicStatus` and `projectCompleted`

Addresses #627

### DIFF
--- a/app/routes/my-projects/archive.js
+++ b/app/routes/my-projects/archive.js
@@ -1,4 +1,15 @@
 import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
 
 export default class MyProjectsArchiveRoute extends Route {
+  @service
+  currentUser;
+
+  @service
+  store;
+
+  async model() {
+    // Use this endpoint for now. This will need to be updated when the backend is finalized.
+    return this.store.query('project', { project_lup_status: 'archive' });
+  }
 }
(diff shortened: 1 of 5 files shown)


app/routes/my-projects/archive.js
```diff
@@ -1,4 +1,15 @@
...
 export default class MyProjectsArchiveRoute extends Route {
+  @service
+  currentUser;
+
+  @service
+  store;
+
+  async model() {
+    // Use this endpoint for now. This will need to be updated when the backend is finalized.
+    return this.store.query('project', { project_lup_status: 'archive' });
```
godfreyyeung: According to the LUP Tab wireframes it looks like we'll need some milestones on the right-hand-side of the proejct cards, so we'll probably also need to "sideload" milestones with the projects. We can do so by adding  the `included` property to the project query:
```
{ project_lup_status: 'upcoming', include: 'milestones' }
```

PR #745, opened 2019-10-03
658 filter milestone lists in upcoming and reviewed tabs

godfreyyeung:
For upcoming and reviewed tab project cards, filters down milestones in milestone lists to predetermined subsets.


app/services/milestone-constants.js
```diff
@@ -0,0 +1,88 @@
...
+    this.BOROUGH_BOARD_REFERRAL,
+  ];
+
+  toReview = [
+    this.COMMUNITY_BOARD_REFERRAL,
+    this.BOROUGH_PRESIDENT_REFERRAL,
+    this.BOROUGH_BOARD_REFERRAL,
+  ];
+
+  reviewed = [
```
hannahkates:
```suggestion
  reviewed = [
    this.COMMUNITY_BOARD_REFERRAL,
    this.BOROUGH_PRESIDENT_REFERRAL,
    this.BOROUGH_BOARD_REFERRAL,
```

hannahkates: ^ need to include these LUP public review milestones too

app/models/project.js
```diff
@@ -54,6 +58,12 @@ export default class ProjectModel extends Model {
     return dcpPlannedstartdate || null;
   }
 
+  @computed('milestones')
+  get tabSpecificMilestones() {
+    const tabSpecificMilestones = this.milestones.filter(milestone => this.milestoneConstants.milestoneListByTabLookup[this.tab].includes(milestone.dcpMilestone));
+    return tabSpecificMilestones;
```
allthesignals: Style nit but I would just return this without creating a new variable name:

```
return this.milestones.filter(milestone => this.milestoneConstants.milestoneListByTabLookup[this.tab].includes(milestone.dcpMilestone));
```

No?

allthesignals: If it's because you want to console log the variable, you can just reference it in a test with whatever value you're looking for

app/services/milestone-constants.js
```diff
@@ -0,0 +1,88 @@
...
+    this.FINAL_LETTER_SENT,
+  ];
+
+  archive = [
+    this.COMMUNITY_BOARD_REFERRAL,
+    this.BOROUGH_PRESIDENT_REFERRAL,
+    this.BOROUGH_BOARD_REFERRAL,
+    this.CPC_PUBLIC_MEETING_VOTE,
+    this.CITY_COUNCIL_REVIEW,
+    this.MAYORAL_VOTE,
```
hannahkates:
```suggestion
    this.MAYORAL_VOTE,
    this.FINAL_LETTER_SENT,
```

hannahkates: I think we should include the final letter for consistency with the "Reviewed" tab

app/services/milestone-constants.js
```diff
@@ -0,0 +1,88 @@
...
+    this.FILED_EAS_REVIEW,
+    this.FINAL_SCOPE_OF_WORK_ISSUED,
+    this.EIS_PUBLIC_SCOPING_MEETING,
+    this.DEIS_PUBLIC_HEARING_HELD,
+    this.FEIS_SUBMITTED_AND_REVIEW,
+    this.NOC_OF_DRAFT_EIS_ISSUED,
+    this.REVIEW_SESSION_CERTIFIED_REFERRED,
+    this.COMMUNITY_BOARD_REFERRAL,
+    this.BOROUGH_PRESIDENT_REFERRAL,
+    this.BOROUGH_BOARD_REFERRAL,
```
allthesignals: Hmmm this works, right? Can't remember if `this` refers to the class scope, or the outer global scope in which the class itself is defined.

app/models/project.js
godfreyyeung: oh woops, yes I was exactly doing it for the console logging and forgot to change it back 😣

app/services/milestone-constants.js
godfreyyeung: Hmm, seems like it will work as long as I don't `new` the service, which I don't think is what Ember does during service injection? 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#Boxing_with_prototype_and_static_methods

app/services/milestone-constants.js
godfreyyeung: okay, will add them. thanks!

app/services/milestone-constants.js
godfreyyeung: will do! 👍

app/services/milestone-constants.js
godfreyyeung: I was wondering the same thing, though

PR #754, opened 2019-10-04
add milestones to archive route

hannahkates:
Updates the archive route to obtain the necessary milestones when the tab is loaded. Now it matches the rest of the tab routes.

### DIFF
--- a/app/routes/my-projects/archive.js
+++ b/app/routes/my-projects/archive.js
@@ -10,6 +10,9 @@ export default class MyProjectsArchiveRoute extends Route {
 
   async model() {
     // Use this endpoint for now. This will need to be updated when the backend is finalized.
-    return this.store.query('project', { project_lup_status: 'archive' });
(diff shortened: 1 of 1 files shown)


app/routes/my-projects/archive.js
```diff
@@ -10,6 +10,9 @@ export default class MyProjectsArchiveRoute extends Route {
 
   async model() {
     // Use this endpoint for now. This will need to be updated when the backend is finalized.
-    return this.store.query('project', { project_lup_status: 'archive' });
+    const archiveProjects = await this.store.query('project', { project_lup_status: 'archive', include: 'actions,milestones,dispositions.action' }, {
+      reload: true,
+    });
+    return archiveProjects;
```
allthesignals: You can also just directly return this value: 
```
return this.store.query('project', { project_lup_status: 'archive', include: 'actions,milestones,dispositions.action' }, {
  reload: true,
});
```

I know the other routes don't do that though :)

PR #757, opened 2019-10-04
738: bp mirage projects

hannahkates:
Address #738 

Changes:
- Adds project milestone scenarios for the BP user
- Sets BP user as the first user

Issues to reconcile:
- The countdown dates on the "upcoming" tab are returning 'invalid' so I think something is wrong with the computed value on the model.
- The countdown dates on the "to review" tab aren't appearing as expected for the project where the BP is assigned both a BP and BB role. I'm not sure how to correctly inject that participant type property onto the project.

### DIFF
--- a/mirage/scenarios/default.js
+++ b/mirage/scenarios/default.js
@@ -1,11 +1,21 @@
 import moment from 'moment';
 
 const NUM_CB_USER_PROJECTS = 8;
-const NUM_BP_USER_PROJECTS = 4;
+const NUM_BP_USER_PROJECTS = 9;
 
 export default function(server) {
-  const seedCBUser = server.create('user', {
+  const seedBPUser = server.create('user', {
     id: 1,
+    email: 'bxbp@planning.nyc.gov',
+    landUseParticipant: 'BXBP',
+  });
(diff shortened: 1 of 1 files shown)


mirage/scenarios/default.js
```diff
@@ -862,30 +860,646 @@ export default function(server) {
...
-  // has reached Public Review but not yet their respective Review Period
   server.create('milestone', 'communityBoardReview', {
     project: seedBPUserProjects[0],
     statuscode: 'In Progress',
-    dcpActualstartdate: moment().subtract(1, 'days'),
-    displayDate: moment().subtract(1, 'days'),
+    dcpActualstartdate: moment().subtract(2, 'days'),
+    displayDate: moment().subtract(2, 'days'),
+    dcpActualenddate: moment().add(28, 'days'),
+    displayDate2: moment().add(28, 'days'),
```
godfreyyeung: @hannahkates   If this first project (project[0]) is 'filed', that means it hasn't reached public review right? 

If public review hasn't started, shouldn't community board review's `dcpActualenddate` not be defined yet?

godfreyyeung:
Instead addressed by #800

Issue #670, opened 2019-09-18
Update error handling for `onConfirm` for hearing form


trbmcginnis:
Currently, the action `onConfirm` that saves the hearing information to the model uses a `try catch` block to handle errors. This needs to be updated and tested.

For more info visit PR 669 --> https://github.com/NYCPlanning/labs-zap-search/pull/669

allthesignals:
- [ ] Simulate an error response in Mirage via the acceptance test
 - [ ] Make sure the try-catch block here is actually needed

PR #766, opened 2019-10-07
670: update error handling for hearing form

trbmcginnis:
First attempt at better error handling for submitting the hearing form. First time trying this out so feel free to edit/comment/suggest something better!

### Changes:
- update error handling on `hearing/add.js` for `.save()` method using`dispositions.save().then`
- if `error` --> show alert message in modal

### Tests:
- update `user-can-fill-out-hearing-form` acceptance test to have check for alert message when 500 error occurs

Addresses #670

### DIFF
--- a/app/controllers/my-projects/project/hearing/add.js
+++ b/app/controllers/my-projects/project/hearing/add.js
@@ -85,28 +85,28 @@ export default class MyProjectsProjectHearingAddController extends Controller {
     const { dispositions } = this.model;
     const allActions = this.get('allActions');
 
-    try {
-      // if user is submitting ONE hearing for ALL actions
-      if (allActions) {
-        const allActionsDispHearingLocation = this.get('dispositionForAllActions.dcpPublichearinglocation');
-        const allActionsDispHearingDate = this.get('dispositionForAllActions.dcpDateofpublichearing');
+    // if user is submitting ONE hearing for ALL actions
+    if (allActions) {
+      const allActionsDispHearingLocation = this.get('dispositionForAllActions.dcpPublichearinglocation');
+      const allActionsDispHearingDate = this.get('dispositionForAllActions.dcpDateofpublichearing');
 
-        // iterate through each disposition on the modal
-        // set hearing location and date on each one with the single value that the user input
-        // this single value is saved on dispositionForAllActions object
-        dispositions.forEach(function(disposition) {
-          disposition.set('dcpPublichearinglocation', allActionsDispHearingLocation);
-          disposition.set('dcpDateofpublichearing', allActionsDispHearingDate);
-        });
-      }
-      await dispositions.save();
+      // iterate through each disposition on the modal
+      // set hearing location and date on each one with the single value that the user input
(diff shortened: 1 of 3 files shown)


tests/acceptance/user-can-fill-out-hearing-form-test.js
```diff
@@ -335,4 +335,37 @@ module('Acceptance | user can save hearing form', function(hooks) {
 
     assert.equal(currentURL(), '/my-projects/4/hearing/done');
   });
+
+  test('if there is a server error when running .save(), user will see error message', async function(assert) {
+    this.server.create('disposition');
+    this.server.patch('/dispositions/:id', { errors: ['server problem'] }, 500); // force mirage to error
```
godfreyyeung: This is super neat! Strangely I can't seem to find docs that talk about mocking these endpoints (`.get`, `.patch`) via `this.server`. Were you able to find any?

I did find this article, though... https://embermap.com/notes/86-testing-mirage-errors-in-your-routes
Also guessing it may have to do with Pretender: 
https://www.ember-cli-mirage.com/docs/api/modules/server~Server#pretender

trbmcginnis: @godfreyyeung I mainly referenced this article: https://embermap.com/notes/86-testing-mirage-errors-in-your-routes

app/controllers/my-projects/project/hearing/add.js
```diff
@@ -85,28 +85,30 @@ export default class MyProjectsProjectHearingAddController extends Controller {
...
+      // this single value is saved on dispositionForAllActions object
+      dispositions.forEach(function(disposition) {
+        disposition.set('dcpPublichearinglocation', allActionsDispHearingLocation);
+        disposition.set('dcpDateofpublichearing', allActionsDispHearingDate);
+      });
     }
+
+    const self = this;
+
+    await dispositions.save().then(function() {
```
allthesignals: I would use an arrow function in line 104 here: `.then(() => {`... the reason being that with arrow functions, we no longer need to use `const self = this` — you can just use `this` inside the arrow function because it inherits the outer scope's `this`. 

See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

allthesignals: Also, because you're `await`ing the promise returned from `dispositions.save()`, you don't need to use `.then` — this is redundant because `await` implicitly "thens" the promise.

Check it out: https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Async_await

tests/acceptance/user-can-fill-out-hearing-form-test.js
allthesignals: yup, `server` is a reference to the Pretender instance, which is what powers Ember Mirage under the hood.

We have a sub to Ember Map if you need the p/w login, just use the labs_dl email (there's no password).

PR #787, opened 2019-10-11
switch Search's default filters to be stage:filed,review

### LINKED ISSUE #679: Default filter on date certified excludes filed projects
The default filter based date certified excludes filed projects (since they don't have certification dates yet). This has been causing a lot of confusion for users and data admins who can't find projects and think they're missing.

**To reproduce:**
- Text search for "greenpoint landing"
- See that only one result appears: P2016K0322, "Greenpoint Landing Parcel 5C - Waterfront ZC, ZA"
- Turn off date filter
- See that there are two additional results now:
  - [not ok] 2019K0219, a new project that hasn't certified yet and should be visible to the public by default https://zap.planning.nyc.gov/projects/2019K0219
  - [ok] P2012K0231, an older project that certified in 2013 so it's expected that it was excluded in the default filter

This is a problem because recently filed projects don't appear by default when those should be of high interest to users.

**Solution:**
After discussing with David and Aline, the best solution is to change the default filter to user Public Status, showing projects where Public Status = "Filed" and projects where Public Status = "In Public Review"

We also need to move the project status filter to be the first on the list.

andycochran:
This PR updates the default filters on the Search view to be the "PROJECT STAGE" instead of "DATE CERTIFIED / REFERRED". Also moves that default filter to the top of the list of filters. 

This is intended to prevent the confusion of users who don't know there's a default filter applied being unable to find projects because they're out of the default date range. 

Closes #679

### DIFF
--- a/app/controllers/query-parameters/show-geography.js
+++ b/app/controllers/query-parameters/show-geography.js
@@ -16,7 +16,7 @@ export const projectParams = new QueryParams({
  * filters it should send to the API.
  */
   'applied-filters': {
-    defaultValue: ['dcp_certifiedreferred'].sort(),
+    defaultValue: ['dcp_publicstatus'].sort(),
     refresh: true,
     serialize(value) {
       return value.toString();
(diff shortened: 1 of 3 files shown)


allthesignals:
Yeah, the tests are going to fail because they're checking that the defaults are working as intended. Will take a look on fixing tests...

andycochran:
@allthesignals Thanks. But you don't have to fix my test. I broke 'em. Unless you you're already working on this, I'll take care of it.

Issue #770, opened 2019-10-08
Combine BB and BP reviews when in-progress on "reviewed" tab


hannahkates:
Combine these two 20 days remaining displays into one. They will always have the same start and end date. It should say "Borough President & and Borough Board Review"
![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/0f840f08-9dae-4a4f-9ca8-62823fb0c017)

allthesignals:
I _think_ this should be fine now that the view should be listing dispositions, because the template will reference the disposition's project's milestones to find this data.

allthesignals:
see reviewed-project-card.hbs

PR #794, opened 2019-10-15
770: combine active bp & bb review on reviewed tab

hannahkates:
Addresses #770 

This PR adds a check on the `reviewed-project-card` component. It checks if the project has multiple in-progress milestones and if those are BP and BB review. If so, it combines the countdown displays to show both reviews together instead of duplicating the same date info for each.

image

### DIFF
--- a/app/components/reviewed-project-card.js
+++ b/app/components/reviewed-project-card.js
@@ -7,15 +7,39 @@ export default class ReviewedProjectCardComponent extends Component {
   @service
   currentUser;
 
-  project = {};
+  project = {
+    reviewedMilestoneActualStartEndDates: [],
+  };
+
+  // Assumes at most two milestones are 'In Progress'
+  @computed('project.reviewedMilestoneActualStartEndDates')
+  get timeDisplayLabel() {
+    const inProgressMilestonesDates = this.project.reviewedMilestoneActualStartEndDates;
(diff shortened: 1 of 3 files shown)


godfreyyeung:
This works, but I would suggest the alternative of doing this through the template. It'll make it cleaner and more explicit. The proposed code hides a lot of the logic within the computed, and makes its purpose less clear. For example, in the template we can write something like: 

```
  <div class="lup-countdown medium-margin-right medium-margin-left">
        {{#each this.timeDisplays as |timeDisplay|}}
            <p class="cert-date">{{timeDisplay.displayName}}</p>
        {{/each}}
            <p class="days">{{this.timeDisplays.firstObject.timeRemaining}}</p>
            <p class="total">of {{this.timeDisplays.firstObject.timeDuration}} days remain</p>
            <p class="end-date">Ends {{moment-format this.timeDisplays.firstObject.dcpActualenddate "M/D/YYYY"}}</p>
  </div>
```

If we stick with the computed, the computed actually no longer needs to return an array because it (theoretically) will always just return one timeDisplay object.

godfreyyeung:
Or actually, we should break down the computed into two separate computeds:
`timeDisplay`  - returns one object based off of `this.project.reviewedMilestoneActualStartEndDates[0]`
`timeDisplayLabel`- returns one string based off of all milestones in `this.project.reviewedMilestoneActualStartEndDates`

Issue #793, opened 2019-10-15
Need to handle invalid dates on milestones


hannahkates:
I think we may be trying to grab a date that doesn't exist. This is a problem on the upcoming tab and reviewed tab.

Upcoming:
![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/0b50543e-5543-44c1-8802-1565a4f7b8fa)
Reviewed:
![image](https://user-images.githubusercontent.com/13967925/66850112-217b7900-ef46-11e9-929d-46afa02c6dc8.png)

hannahkates:
Notes while troubleshooting:

- I updated the logic on the "reviewed" countdown to show the planned end date if the actual is NULL. That solved the ugly NaN response, but uncovered another problem. If the milestone doesn't get closed and doesn't get a real actual end date (data quality problem), we end up with negative numbers like this screenshot. We need to figure out if certain milestones shouldn't be expected to have actual end dates.

![image](https://user-images.githubusercontent.com/13967925/67040103-c5a12380-f0f0-11e9-8d40-b1d2db20ee16.png)

PR #811, opened 2019-10-18
793: dont show invalid dates, show planned where appropriate

hannahkates:
This PR solves a few issues related to milestone displays:

- Incorrect milestone name labels. We should be using displayName instead of milestonename. displayName is a recoded attribute that provides more user friendly labels than CRM's.

- (Addresses #793) Milestone displayDate is calculated in the API using previously determined logic for which dates DCP feels comfortable showing the public (based on confidence/reliability). The front-end was assuming that displayDate will always be provided, but some milestones that haven't started yet purposefully don't have a displayDate. This was causing a bunch of "invalid date" dates being displayed. I've added logic to the "upcoming" and "reviewed" templates to not show dates if they're not provided.

- (Addresses #793) dcpActualenddate returned in the project model's computed `reviewedMilestoneActualStartEndDates()` property assumed that dcpActualenddate would always be available for an "In Progress" milestone. That's not true because some milestones don't have a known, mandated duration. The incorrect assumption was causing NaN countdowns for some "In Progress" milestones. I added logic in the project model to set dcpActualenddate = dcpPlannedcompletiondate when dcpActualenddate is NULL. This is potentially confusing because it's being called "actual" when it's planned. Curious to get @NYCPlanning/engineering 's thoughts on that.

I uncovered [several other data display concerns](https://github.com/NYCPlanning/labs-zap-search/issues/793#issuecomment-543321433) while sleuthing through this that I'll create separate issues for.

### DIFF
--- a/app/components/reviewed-project-card.js
+++ b/app/components/reviewed-project-card.js
@@ -8,13 +8,13 @@ export default class ReviewedProjectCardComponent extends Component {
   currentUser;
 
   assignment = {
-    reviewedMilestoneActualStartEndDates: [],
+    reviewedMilestoneDates: [],
   };
 
   // Assumes at most two milestones are 'In Progress'
-  @computed('assignment.reviewedMilestoneActualStartEndDates')
+  @computed('assignment.reviewedMilestoneDates')
   get timeDisplayLabel() {
-    const inProgressMilestonesDates = this.assignment.reviewedMilestoneActualStartEndDates;
+    const inProgressMilestonesDates = this.assignment.reviewedMilestoneDates;
     if (inProgressMilestonesDates.length > 1 && inProgressMilestonesDates[0] && inProgressMilestonesDates[1]) {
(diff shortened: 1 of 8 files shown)


app/models/project.js
```diff
@@ -117,7 +117,7 @@ export default class ProjectModel extends Model {
       milestone: milestone.dcpMilestone,
       displayName: milestone.displayName,
       dcpActualstartdate: milestone.dcpActualstartdate,
-      dcpActualenddate: milestone.dcpActualenddate,
+      dcpActualenddate: ((milestone.dcpActualenddate) ? milestone.dcpActualenddate : milestone.dcpPlannedcompletiondate),
```
andycochran: This seems like it should be template logic, not model logic. Setting the actual date to equal a planned date could cause trouble.

hannahkates: I moved this logic to the template and controller so it's easier to see when the dates are actual vs planned.

Issue #771, opened 2019-10-08
Correct display for planned start on "upcoming" tab


hannahkates:
The executive office has agreed to the approach of showing “Within 30 days”, “More than 30 days” on the "upcoming" tab instead of a specific estimated start date.

Instead of displaying the content shown in screenshot below, we should display...

**Before certification has occurred (all users)**
i.e. `dcpActualenddate` for the cert milestone IS NULL

- top line: "Estimated Certification Date" 
- 2nd big line: "Within 30 days" (calculated using `dcpPlannedcompletiondate` for the cert milestone)
- (nix 3rd line)

**Once certification has occurred (only relevant for BP users but can be applied to all users):**
~
- top line: "Date Certified" 
- 2nd big line: Actual date of certification (using `dcpActualenddate` for the cert milestone)
- (nix 3rd line)
~
![image](https://user-images.githubusercontent.com/13967925/66937078-d0828800-f00c-11e9-8711-91ea1a144ecb.png)

PR #812, opened 2019-10-18
771: fuzzy public review start date

hannahkates:
Prior to certification, the exec office wants to show fuzzy count down dates for when public review is expected to begin. This PR implements that by 
- Fixing the computed `upcomingMilestonePlannedStartDate` property on the project model. The API isn't providing any attributes for `dcpPublicstatusSimp`, so filed projects were being treated like certified projects. We need to use `dcpPublicstatus` instead.
- Adding a relative date computed value to the `upcoming-project-card` component

@andycochran the font size hierarchy looks a little off now that we're displaying text instead of purely dates. I started with using real words instead of "<" but the font was too big. Can you take a look at the design/text and see if you think it should be tweaked?

Addresses #771

### DIFF
--- a/app/components/upcoming-project-card.js
+++ b/app/components/upcoming-project-card.js
@@ -1,6 +1,7 @@
 import Component from '@ember/component';
-import { action } from '@ember/object';
+import { computed, action } from '@ember/object';
 import { inject as service } from '@ember/service';
+import moment from 'moment';
 
 export default class UpcomingProjectCardComponent extends Component {
   @service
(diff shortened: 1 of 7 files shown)


andycochran:
I've made the styles of the timing column more consistent across the tabs.

godfreyyeung:
I can help rework this to be assignments-based

app/components/upcoming-project-card.js
```diff
@@ -10,6 +11,14 @@ export default class UpcomingProjectCardComponent extends Component {
 
   showPopup = false;
 
+  @computed('project.publicReviewPlannedStartDate')
```
godfreyyeung: @hannahkates should be `@computed('assignment.publicReviewPlannedStartDate')` here

app/models/assignment.js
```diff
@@ -55,7 +55,7 @@ export default class AssignmentModel extends Model {
   @computed('tab', 'dcpLupteammemberrole', 'project.{milestones,dcpPublicstatusSimp}')
```
godfreyyeung: @hannahkates this needs to be ` @computed('tab', 'dcpLupteammemberrole', 'project.{milestones,dcpPublicstatus}')`

PR #817, opened 2019-10-21
816 fixup done pages, other small patches

godfreyyeung:
- update recommmendation/done and hearing/done to access project through assignment
- add tests to make sure navigation away from done pages work
- bugfix: in show-project controller, ensure currentUser.user is instantiated before computing `isUserAssignedToProject` 
- get rid of unnecessary `.get()`

### DIFF
--- a/app/controllers/show-project.js
+++ b/app/controllers/show-project.js
@@ -41,7 +41,7 @@ export default class ShowProjectController extends Controller {
   get isUserAssignedToProject() {
     // user is from currentUser Service
     // currentUser Service injected in show-project route on setUpController
-    if (this.session.isAuthenticated) {
+    if (this.user && this.session.isAuthenticated) {
       const userProjectIds = this.user.assignments.map(({ project }) => project.id);
 
       return userProjectIds.includes(this.model.id);
--- a/tests/acceptance/user-can-navigate-away-from-done-pages-test.js
+++ b/tests/acceptance/user-can-navigate-away-from-done-pages-test.js
@@ -1,0 +1,177 @@
+import { module, test } from 'qunit';
+import {
+  click,
+  visit,
+  currentURL,
+} from '@ember/test-helpers';
+import { setupApplicationTest } from 'ember-qunit';
+import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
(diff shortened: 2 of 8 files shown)


app/controllers/show-project.js
```diff
@@ -41,7 +41,7 @@ export default class ShowProjectController extends Controller {
   get isUserAssignedToProject() {
     // user is from currentUser Service
     // currentUser Service injected in show-project route on setUpController
-    if (this.session.isAuthenticated) {
+    if (this.user && this.session.isAuthenticated) {
```
allthesignals: fair

package.json
```diff
@@ -15,7 +15,7 @@
     "lint": "./node_modules/.bin/eslint ./*.js app config lib server tests",
     "lint:hbs": "ember-template-lint .",
     "start": "ember serve",
-    "test": "ember test"
+    "test": "ember test --silent --reporter dot"
```
allthesignals: did this work? lemme know, great if it did! what does each statement do?

tests/acceptance/user-can-navigate-away-from-done-pages-test.js
```diff
@@ -0,0 +1,177 @@
...
+
+    await authenticateSession({
+      id: 1,
+    });
+
+    await visit('/my-projects/1/hearing/done');
+    assert.equal(currentURL(), '/my-projects/1/hearing/done');
+
+    await click('[data-test-button="back-to-project-profile"]');
+    assert.equal(currentURL(), '/projects/2');
```
allthesignals: ~why is it projects/2?~

nvm!!!! my-projects is really my-assignments ;)

app/controllers/show-project.js
godfreyyeung: Yah, it was throwing "cannot access undefined" for  `this.user.assignments` for when the CP was computing before `this.user` was available.

package.json
godfreyyeung: `--silent` supposedly supresses top-level warnings and `to silence unwanted output`, but it doesn't necessarily do so for warnings stemming from addons and plugins? `--reporter dot` represents tests passing/failing using this dot notation:

Screen Shot 2019-10-21 at 5 54 50 PM

I'm thinking I should actually change it back for now, since it's useful to have it be a little verbose.

Issue #815, opened 2019-10-21
Clarify which (deduped) hearings list will apply to project's dispositions and which apply to assignment dispositions


godfreyyeung:
We need to decide whether the hearings lists in each of the following places should show ALL hearings in a project, or show only the assignment hearings.  

- show-project  (project profile) page "To Review section"
- show-project  (project profile) page milestones listing
- upcoming-project-card
- to-review-project-card
- reviewed-project-card
- hearing form confirmation modal
- recommendation form quorum question
- recommendation form confirmation modal

Basically,  this will boil down to either passing `project.dispositions` or `assignment.dispositions` into components like `deduped-hearings-list

trbmcginnis:
Here's what I'm thinking:

*On any milestones list that's shown on tabs or on `show-project.js`:
- only show hearings that are associated with that project's milestone
- do this through `project-milestone` component, use lookup table to compare dispositions and milestones and create a new array of dispositions that ONLY has the dispositions associated with that project's milestone

*On Recommendation and Hearing Form
- only show hearings that are associated with that user's current project --> `assignment.dispositions`

*`to-review` tab:
- this is where we need design help --> should we only show the hearings that a user just submitted for that project OR should we show ALL hearings associated with that project that have been submitted
- if show ALL hearings associated with a project, then denote which user submitted each hearing (so it would be `project.dispositions`)

godfreyyeung:
The design decisions on this may also affect whether hearings computed are moved from the project model to the assignments model, or if both models will have the hearings computed.

godfreyyeung:
show-project (project profile) page "To Review section" -- assignment.disposition
show-project (project profile) page milestones listing -- project.disposition . 
upcoming-project-card -- project.dispositions
to-review-project-card .-- assignment.dispositions
reviewed-project-card -- project.dispositions
hearing form confirmation modal -- assignment.disposition
recommendation form quorum question -- assignment.disposition 
recommendation form confirmation modal -- assignment.disposition

PR #827, opened 2019-10-22
815: refactor the implementation of deduped-hearings-list

trbmcginnis:
### Changes:
- refactor `deduped-hearings-list` to create a new object rather than set properties directly on the model
- Add a component called `hearings-list-for-milestones-list` that displays a deduped hearings list based on the current milestone in the list -- this is rendered on `upcoming-project-card`, `archive-project-card`, and `reviewed-project-card`
- Remove option to submit/waive hearings directly on `show-project` and instead there's just a button that links to `to-review` route
- removal of current user's hearings list from `upcoming-project-card` because they will now show up on the milestones instead
- updates on routes and templates to match assignments refactor

### To-Do:
- create a hearings waived option on the milestones list
- Include scenario where there might be more than one Community Board for a project
- further testing of new component `hearings-list-for-milestones-list`

Addresses #815


app/components/deduped-hearings-list.js
```diff
@@ -32,28 +32,32 @@ export default class DedupedHearingsListComponent extends Component {
...
+
+    this.dispositions.forEach(function(currentDisp) {
+      newDispositionsArray.push(
+        {
+          disposition: currentDisp,
+          hearingActions: [currentDisp.action],
+          duplicateDisps: [currentDisp],
+        },
+      );
     });
```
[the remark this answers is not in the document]
allthesignals: One alternative here would be to use `map`:

```
const newDispositionsArray = this.dispositions.map(disposition => ({
  disposition: disposition,
  hearingActions: [disposition.action],
  duplicateDisps: [disposition],
});
```

`map` returns a new array entirely, so it'd save you the step of declaring `newDispositionsArray` in the upper scope, looping, pushing, and so on.

app/components/hearings-list-for-milestones-list.js
```diff
@@ -0,0 +1,66 @@
...
+      // milestoneParticipantReviewLookup[participantType] = e.g. `Community Board Review`
+      if (milestone.displayName === milestoneParticipantReviewLookup[participantType]) {
+        currentMilestoneDispositions.push(disposition);
+      }
+    });
+
+    // if dispositions have been pushed to the array (its length is greater than 1)
+    // return arrray -- otherwise return null
+    if (currentMilestoneDispositions.length < 1) {
+      return null;
```
allthesignals: why return null? an empty array would prevent situations where calling `.map` or `.forEach` on this computed results in an error.

app/components/hearings-list-for-milestones-list.js
```diff
@@ -0,0 +1,66 @@
...
+    // current milestone's displayName, push that disposition into the new
+    // currentMilestoneDispositions array
+    dispositions.forEach(function(disposition) {
+      const participantType = disposition.get('assignment.dcpLupteammemberrole');
+
+      // milestoneParticipantReviewLookup[participantType] = e.g. `Community Board Review`
+      if (milestone.displayName === milestoneParticipantReviewLookup[participantType]) {
+        currentMilestoneDispositions.push(disposition);
+      }
+    });
```
allthesignals: Similar to deduped-hearings-list, but it looks like you want a `filter` method here, which returns a new array also:

```javascript
const currentMilestoneDispositions = dispositions.filter((disposition) => {
 // do your filtering logic here, with the if cases
  const participantType = disposition.get('assignment.dcpLupteammemberrole');

  return milestone.displayName === milestoneParticipantReviewLookup[participantType];
});
```

You could even just `return` the call to `dispositions.filter` :)

app/components/hearings-list-for-milestones-list.js
```diff
@@ -0,0 +1,66 @@
...
+
+    // if dispositions have been pushed to the array (its length is greater than 1)
+    // return arrray -- otherwise return null
+    if (currentMilestoneDispositions.length < 1) {
+      return null;
+    } return currentMilestoneDispositions;
+  }
+
+  // An array of disposition models that match the current milestone
```
allthesignals: Is this comment still true?

app/components/hearings-list-for-milestones-list.js
```diff
@@ -0,0 +1,66 @@
...
+  @computed('milestone.milestonename')
+  get currentMilestoneUserType() {
+    const milestoneName = this.get('milestone.milestonename');
+    return milestoneName.replace(' Review', '');
+  }
```
allthesignals: Do you think this computed could go on the milestone model itself? it's only dependency is the `milestone` prop.

app/components/hearings-list-for-milestones-list.js
```diff
@@ -0,0 +1,66 @@
...
+
+
+  // if each dcpPublichearinglocation and dcpDateofpublichearing properties are filled in currentMilestoneDispositions array,
+  // then hearings have been submitted for that project
+  @computed('currentMilestoneDispositions.@each.{dcpPublichearinglocation,dcpDateofpublichearing}')
+  get hearingsSubmitted() {
+    const dispositions = this.get('currentMilestoneDispositions');
```
allthesignals: Reiterating my comment above, if/when `currentMilestoneDispositions` returns null, the next line will break.

app/components/hearings-list-for-milestones-list.js
```diff
@@ -0,0 +1,66 @@
...
+    // array of hearing locations
+    const dispositionHearingLocations = dispositions.map(disp => `${disp.dcpPublichearinglocation}`);
+    // array of hearing dates
+    const dispositionHearingDates = dispositions.map(disp => disp.dcpDateofpublichearing);
+    // hearingsSubmittedForProject checks whether each item in array is truthy
+    const arraysFilled = dispositionHearingLocations.length > 0 && dispositionHearingDates.length > 0;
+    const hearingsSubmittedForProject = arraysFilled && dispositionHearingLocations.every(location => !!location) && dispositionHearingDates.every(date => !!date);
+    return hearingsSubmittedForProject;
```
allthesignals: ~hmm what's the requirement? you want to make sure that _all_ hearing locations and dates are filled out? it looks like this would only check if at least _some_ of them are, not _all_.~

ah okay, so you're using `every` to check — great! I think you could simplify this _a lot_ by always returning an array from the `currentMilestoneDispositions` computed prop. When I'm in these situations, I always try to do everything I can to make sure that the computed I'm writing always consistently returns the same primitive. This isn't _always_ possible, but it does help with debugging. Because it looks like your `arraysFilled` here is checking that they're arrays? Not quite sure... At any rate, you could probably just do this:

```javascript
return this.currentMilestoneDispositions.every((disposition) => {
  const { dcpPublichearinglocation, dcpDateofpublichearing } = disposition;

  // no need to coerce to boolean with double bang—these are ember model attrs
  //and are expected to be strings/null, which are implicitly coercible to boolean.
  return dcpPublichearinglocation && dcpPublichearinglocation;
});
```

you could even get extra fancy with destructuring out of the function signature:
```javascript
return this.currentMilestoneDispositions
  .every(({ dcpPublichearinglocation, dcpPublichearinglocation }) => dcpPublichearinglocation && dcpPublichearinglocation);
```

app/models/assignment.js
```diff
@@ -34,6 +34,47 @@ export default class AssignmentModel extends Model {
...
+  get hearingsSubmitted() {
+    const dispositions = this.get('dispositions');
+    // array of hearing locations
+    const dispositionHearingLocations = dispositions.map(disp => `${disp.dcpPublichearinglocation}`);
+    // array of hearing dates
+    const dispositionHearingDates = dispositions.map(disp => disp.dcpDateofpublichearing);
+    // hearingsSubmittedForProject checks whether each item in array is truthy
+    const hearingsSubmittedForProject = dispositionHearingLocations.every(location => !!location) && dispositionHearingDates.every(date => !!date);
+    return hearingsSubmittedForProject;
+  }
```
allthesignals: This looks familiar... seems like we need this kind of validation check on multiple things? I like that it's here on the model.

app/models/assignment.js
```diff
@@ -34,6 +34,47 @@ export default class AssignmentModel extends Model {
...
+  // then hearings have been waived
+  @computed('dispositions.@each.dcpPublichearinglocation')
+  get hearingsWaived() {
+    const dispositions = this.get('dispositions');
+
+    // array of hearing locations
+    const dispositionHearingLocations = dispositions.map(disp => disp.dcpPublichearinglocation);
+    // each location field equal to 'waived'
+
+    return dispositionHearingLocations.every(location => location === 'waived');
```
allthesignals: nice!

app/models/assignment.js
```diff
@@ -34,6 +34,47 @@ export default class AssignmentModel extends Model {
...
+  }
+
+  @computed('hearingsSubmitted', 'hearingsWaived')
+  get hearingsSubmittedOrWaived() {
+    const hearingsSubmitted = this.get('hearingsSubmitted');
+    const hearingsWaived = this.get('hearingsWaived');
+    return !!hearingsSubmitted || !!hearingsWaived;
+  }
```
allthesignals: We know that both hearingsSubmitted and hearingsWaived return booleans, so no need to coerce to boolean with the double bang. I would just simplify:

```
return this.hearingsSubmitted || this.hearingsWaived
```

It's shorter, but there's also much less noise.

app/routes/my-projects/archive.js
```diff
@@ -12,7 +12,7 @@ export default class MyProjectsArchiveRoute extends Route {
     // Use this endpoint for now. This will need to be updated when the backend is finalized.
     return this.store.query('assignment', {
       tab: 'archive',
-      include: 'project.milestones,project.dispositions,project.actions',
+      include: 'project.milestones,project.dispositions,project.actions,dispositions,dispositions.action',
```
allthesignals: thanks!

app/routes/my-projects/to-review.js
```diff
@@ -7,4 +7,13 @@ export default class MyProjectsToReviewRoute extends Route {
...
   store;
+
+  async model() {
+    return this.store.query('assignment', {
+      tab: 'to-review',
+      include: 'project.milestones,project.dispositions,project.actions,dispositions,dispositions.action',
+    }, {
+      reload: true,
+    });
+  }
```
allthesignals: For demo tomorrow we need to find a quick fix around this... hard to explain over comments

tests/integration/components/hearings-list-for-milestones-list-test.js
```diff
@@ -0,0 +1,22 @@
+import { module, skip } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import hbs from 'htmlbars-inline-precompile';
+
+module('Integration | Component | hearings-list-for-milestones-list', function(hooks) {
+  setupRenderingTest(hooks);
+
+  skip('it renders', async function(assert) {
```
allthesignals: 🤔

allthesignals: Hmm

trbmcginnis: Did this for the sake of time considering the demo is tomorrow and it was already 7pm -- was going to fix in another refactor where I could focus on creating a more detailed integration test like some of the other components. Currently have an acceptance test that checks for this component's rendering in all three tabs

app/routes/my-projects/to-review.js
trbmcginnis: Okay let's chat tomorrow

app/models/assignment.js
trbmcginnis: Will do

app/models/assignment.js
trbmcginnis: Are you referring to seeing it also on the new hearings-list-on-milestones-list component?  `hearingSubmitted` here checks for `assignment.dispositions` whereas `hearingsSubmitted` in the hearings-list-on-milestones-list checks `project.dispositions` associated with a particular milestone -- so would be validating two different things. Maybe I can move `hearingsSubmitted` that's on the hearings-list-on-milestone-list component to the `milestone` model ?

app/components/hearings-list-for-milestones-list.js
trbmcginnis: Okay cool, will explore :)

app/components/hearings-list-for-milestones-list.js
trbmcginnis: Okay -- will change

app/components/hearings-list-for-milestones-list.js
trbmcginnis: yep!

app/components/hearings-list-for-milestones-list.js
trbmcginnis: Looks like it got shuffled around and is now in the wrong place -- will remove

app/components/hearings-list-for-milestones-list.js
trbmcginnis: Sounds good -- will change

app/components/hearings-list-for-milestones-list.js
trbmcginnis: will do!

app/components/deduped-hearings-list.js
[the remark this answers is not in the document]
trbmcginnis: ooh cool, okay will try it out :)

PR #835, opened 2019-10-23
771 patch condition for displaying fuzzy public review start date on upcoming project card

godfreyyeung:
- refactor assignment 'upcoming' date CPs to have less dependencies, and to handle undefined `dcpPublicStatusSimp`
- fix upcoming project card test

### DIFF
--- a/app/models/assignment.js
+++ b/app/models/assignment.js
@@ -86,23 +86,13 @@ export default class AssignmentModel extends Model {
     return this.project.get('milestones').filter(milestone => this.milestoneConstants.milestoneListByTabLookup[this.tab].includes(milestone.dcpMilestone));
   }
 
-  // The following <tab>MilestoneActual<Start/End>
-  // date fields represent different dates from different milestones
-  // depending on `tab`:
-
-  // If `tab` is `upcoming`...
-  // if the project is in public review, this field is used to display the participant's review planned start date.
-  // else this field returns null
-  @computed('tab', 'dcpLupteammemberrole', 'project.{milestones,dcpPublicstatus}')
+  // This field is used to display the participant's review planned start date.
+  // If not found, returns null
+  @computed('tab', 'dcpLupteammemberrole', 'project.milestones')
   get upcomingMilestonePlannedStartDate() {
-    if (this.tab === 'upcoming') {
-      if (this.project.dcpPublicstatus !== 'Filed') {
-        const participantMilestoneId = MILESTONE_ID_LOOKUP[this.dcpLupteammemberrole];
-        const participantReviewMilestone = this.project.get('milestones').find(milestone => milestone.dcpMilestone === participantMilestoneId);
-        return participantReviewMilestone ? participantReviewMilestone.dcpPlannedstartdate : null;
-      }
-    }
-    return null;
+    const participantMilestoneId = MILESTONE_ID_LOOKUP[this.dcpLupteammemberrole];
+    const participantReviewMilestone = this.project.get('milestones').find(milestone => milestone.dcpMilestone === participantMilestoneId);
+    return participantReviewMilestone ? participantReviewMilestone.dcpPlannedstartdate : null;
   }
 
   // If `tab` is to-review', these start/end dates are derived from
--- a/app/templates/components/upcoming-project-card.hbs
+++ b/app/templates/components/upcoming-project-card.hbs
@@ -5,23 +5,23 @@
   <div class="grid-x grid-x-small-gutters">
     <div class="cell large-4 xlarge-5">
       <h3 class="tiny-margin-bottom">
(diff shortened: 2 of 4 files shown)


app/models/assignment.js
```diff
@@ -86,23 +86,13 @@ export default class AssignmentModel extends Model {
...
-      if (this.project.dcpPublicstatus !== 'Filed') {
-        const participantMilestoneId = MILESTONE_ID_LOOKUP[this.dcpLupteammemberrole];
-        const participantReviewMilestone = this.project.get('milestones').find(milestone => milestone.dcpMilestone === participantMilestoneId);
-        return participantReviewMilestone ? participantReviewMilestone.dcpPlannedstartdate : null;
-      }
-    }
-    return null;
+    const participantMilestoneId = MILESTONE_ID_LOOKUP[this.dcpLupteammemberrole];
+    const participantReviewMilestone = this.project.get('milestones').find(milestone => milestone.dcpMilestone === participantMilestoneId);
+    return participantReviewMilestone ? participantReviewMilestone.dcpPlannedstartdate : null;
```
allthesignals: I wonder if this computed could just return the result of line 94. Then it's the responsibility of the caller to pull out whatever information they want. just an idea

app/templates/components/upcoming-project-card.hbs
```diff
@@ -67,12 +67,12 @@
               <small>Opt Out</small>
             </button>
 
-            {{waive-hearings-popup assignment=assignment showPopup=showPopup}}
+            {{waive-hearings-popup assignment=this.assignment showPopup=showPopup}}
```
allthesignals: thanks for adding `this`!

mirage/factories/project.js
```diff
@@ -10,6 +10,17 @@ export default Factory.extend({
   //   return faker.random.uuid();
   // },
 
+  // dcpPublicStatus and dcpPublicstatusSimp should not receive random mock data
+  // because their values are critical to expressing project state.
+
```
allthesignals: we could turn it into a trait for later but agree it should not be randomized

tests/acceptance/upcoming-project-cards-renders-test.js
```diff
@@ -33,13 +33,15 @@ module('Acceptance | upcoming project cards renders', function(hooks) {
         this.server.create('assignment', {
           id: 1,
           tab: 'upcoming',
-          dcpPublicstatus: 'Filed',
```
allthesignals: i like having this here though because it makes it explicit to future test readers that this is testing for behavior that's downstream of the value of `dcpPublicstatus`, no?

allthesignals: oh wait it's not a real field is it? I don't even think the API returns a dcpPublicstatus... just a dcpPublicstatusSimp

tests/acceptance/upcoming-project-cards-renders-test.js
```diff
@@ -33,13 +33,15 @@ module('Acceptance | upcoming project cards renders', function(hooks) {
...
           dcpLupteammemberrole: 'BP',
           dispositions: [],
           project: this.server.create('project', {
-            milestones: [this.server.create('milestone', 'communityBoardReview', {
-              dcpPlannedstartdate: moment().add(32, 'days'),
-            })],
+            dcpPublicStatusSimp: 'Filed',
```
allthesignals: this is camel case PublicStatus—is that right?

tests/acceptance/upcoming-project-cards-renders-test.js
```diff
@@ -123,4 +125,38 @@ module('Acceptance | upcoming project cards renders', function(hooks) {
...
+    await authenticateSession({
+      id: 1,
+    });
+
+    await visit('/my-projects/upcoming');
+
+    const timeRemainingValue = find('[data-test-fuzzy-time-remaining]').textContent.trim();
+    assert.equal(timeRemainingValue, '< 30 days', 'Time remaining displays less than 30 days');
+    assert.notOk(find('[data-test-time-remaining]'));
+  });
```
allthesignals: great, very clear test! this is pro

tests/acceptance/upcoming-project-cards-renders-test.js
godfreyyeung: Good catch!! 🙇 It shouldn't be camel cased! THANK YOU!

tests/acceptance/upcoming-project-cards-renders-test.js
godfreyyeung: @allthesignals API does return `dcpPublicstatus`, but on staging it was returning `null` or `""`, so Hannah and I decided to revert it back to `dcpPublicstatusSimp`

app/models/assignment.js
godfreyyeung: Hmm, I thought about it but think for now I'll leave it to this component to pull out `dcpPlannedstartdate`, so that the template doesn't have to do an extra check for `undefined` in case the CP can't find the participant milestone.

Issue #686, opened 2019-09-20
why is this async?


allthesignals:
why is this async?

_Originally posted by @allthesignals in https://github.com/NYCPlanning/labs-zap-search/pull/685_

PR #846, opened 2019-10-24
823 wrap moment-format helpers in custom guard

godfreyyeung:
Note that because it seems the `~` whitespace trim operator within a component doesn't work outside of  that component, it's up to the consuming template to trim out whitespace around the `<DisplayDate />` component.

Also takes care of #686

### DIFF
--- a/app/templates/components/archive-project-milestone-list-item.hbs
+++ b/app/templates/components/archive-project-milestone-list-item.hbs
@@ -27,7 +27,7 @@
       {{/each}}
       <br>
       <small class="display-block">
-        Recommendation submitted&nbsp;{{~moment-format this.anyCommunityBoardDispositionDateReceived "MM/D/YYYY"~}}
+        Recommendation submitted&nbsp;<DateDisplay @date={{this.anyCommunityBoardDispositionDateReceived}} />
       </small>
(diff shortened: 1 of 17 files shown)


app/templates/components/archive-project-milestone-list-item.hbs
```diff
@@ -27,7 +27,7 @@
       {{/each}}
       <br>
       <small class="display-block">
-        Recommendation submitted&nbsp;{{~moment-format this.anyCommunityBoardDispositionDateReceived "MM/D/YYYY"~}}
+        Recommendation submitted&nbsp;<DateDisplay @date={{this.anyCommunityBoardDispositionDateReceived}} @outputFormat='MM/D/YYYY' />
```
hannahkates: Awesome, thanks for figuring this out @godfreyyeung. Question to understand: why is it necessary to specify @outputFormat="MM/D/YYYY" when that's the preferred format? I figured that was being set as the default outputFormat in line 13 of the new component, so you'd only need to specific the @outputFormat in the template if it was overriding the default of "MM/D/YYYY".

godfreyyeung: Oh yeah good catch! that's true in this case it's not necessary. I was mass-transforming each `moment-format` to `DisplayDate`. I'll update

PR #852, opened 2019-10-28
768 milestones display upcoming

allthesignals:
This PR provides an abridged list of milestones intended for display in the upcoming tab

### DIFF
--- a/app/models/assignment.js
+++ b/app/models/assignment.js
@@ -6,15 +6,6 @@ const {
   Model, belongsTo, hasMany, attr,
 } = DS;
 
-const COMMUNITY_BOARD_REFERRAL_MILESTONE_ID = '923beec4-dad0-e711-8116-1458d04e2fb8';
-const BOROUGH_PRESIDENT_REFERRAL_MILESTONE_ID = '943beec4-dad0-e711-8116-1458d04e2fb8';
-const BOROUGH_BOARD_REFERRAL_MILESTONE_ID = '963beec4-dad0-e711-8116-1458d04e2fb8';
-const MILESTONE_ID_LOOKUP = {
-  CB: COMMUNITY_BOARD_REFERRAL_MILESTONE_ID,
-  BP: BOROUGH_PRESIDENT_REFERRAL_MILESTONE_ID,
-  BB: BOROUGH_BOARD_REFERRAL_MILESTONE_ID,
-};
-
 export default class AssignmentModel extends Model {
   @service
   milestoneConstants;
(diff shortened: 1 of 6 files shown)


allthesignals:
@godfreyyeung ooh good catch. Let me add that. Thank you.

app/models/assignment.js
```diff
@@ -84,12 +75,57 @@ export default class AssignmentModel extends Model {
...
+        return this.milestones;
+    }
+  }
+
   // This field is used to display the participant's review planned start date.
   // If not found, returns null
   @computed('tab', 'dcpLupteammemberrole', 'project.milestones')
   get upcomingMilestonePlannedStartDate() {
-    const participantMilestoneId = MILESTONE_ID_LOOKUP[this.dcpLupteammemberrole];
+    const participantMilestoneId = this.milestoneConstants.referralIdentifierByAcronymLookup[this.dcpLupteammemberrole];
```
godfreyyeung: :star_struck:

app/models/assignment.js
```diff
@@ -84,12 +75,57 @@ export default class AssignmentModel extends Model {
...
+
+  @computed('project.milestones')
+  get lastCompletedMilestone() {
+    const completedMilestones = this.project.get('milestones').filterBy('statuscode', 'Completed');
+
+    return completedMilestones[completedMilestones.length - 1] || null;
+  }
+
+  // abridged view of milestones typically used in upcoming tab
+  @computed('tab', 'project.milestones')
```
trbmcginnis: Could you add a comment saying what's supposed to show up on the upcoming tab? That way if someone goes in to change the logic they know exactly what's supposed to show up. e.g. "the upcoming tab should show 1) milestones that blah blah, 2) milestones that blah blah etc."

mirage/factories/milestone.js
```diff
@@ -12,6 +13,24 @@ export default Factory.extend({
   dcpMilestoneoutcome: null,
   milestoneLinks: [],
 
+  isCompleted: trait({
```
godfreyyeung: All milestones that use the `isCompleted` trait would have the same `dcpActualstartdate`, and same for all milestones that use `isInProgress` or `isNotStarted`, which doesn't match real life? (a milestone's end date should be after the previous milestone's end date) Thinking that this might also affect/skew tests..

In the future maybe we should have fixtures for `before public review`, `in public review`  `in participant review` and `post review`  milestone lists. Each fixture would contain the entire list of milestones,  the milestone dates would be sequential, and we would assign `statuscode`s appropriate to the fixture. 

It'll be easier to explain in person

PR #862, opened 2019-10-29
new mirage scenarios for cb and bp users

hannahkates:
Changes:
- New mirage CB scenario file for usability testing!
- New mirage BP scenario file for usability testing! (help uncover a few little bugs I'll create issues for)
- Skips auth step so it's easier to review work when running the app locally
- Fixes the hearingsWaived computed value on the assignment model. We were getting back an empty array which the app was incorrectly thinking meant the hearings had been waived.
- Adds conditional date display on the archive tab, only showing actual end dates when available
- Updates fuzzy upcoming start date text

### DIFF
--- a/app/templates/components/archive-project-card.hbs
+++ b/app/templates/components/archive-project-card.hbs
@@ -15,7 +15,11 @@
       <div class="text-center medium-margin-right medium-margin-left">
         <p class="lead small-margin-bottom">
           <strong>
-            {{this.assignment.project.dcpPublicstatusSimp}}
+            {{#if (eq this.assignment.project.dcpPublicstatus 'Withdrawn/Terminated/Disapproved')}}
+              Withdrawn, Terminated, or Disapproved
+            {{else}}
+              {{this.assignment.project.dcpPublicstatus}}
+            {{/if}}
             <DateDisplay
               @date={{this.assignment.project.dcpProjectcompleted}}
               @outputFormat="M/D/YYYY"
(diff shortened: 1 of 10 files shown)


app/templates/components/archive-project-card.hbs
```diff
@@ -15,7 +15,11 @@
       <div class="text-center medium-margin-right medium-margin-left">
         <p class="lead small-margin-bottom">
           <strong>
-            {{this.assignment.project.dcpPublicstatusSimp}}
+            {{#if (eq this.assignment.project.dcpPublicstatus 'Withdrawn/Terminated/Disapproved')}}
+              Withdrawn, Terminated, or Disapproved
+            {{else}}
+              {{this.assignment.project.dcpPublicstatus}}
+            {{/if}}
```
allthesignals: At a later point we'll want to come up with a metadata system that maps domain values to presentable ones

app/templates/components/archive-project-milestone-list-item.hbs
```diff
@@ -53,11 +53,13 @@
...
-          @date={{this.milestone.dcpActualenddate}}
-        />
-      </small>
+      {{#if this.milestone.dcpActualenddate}}
+        <small class="display-block">
+          <DateDisplay
+            @date={{this.milestone.dcpActualenddate}}
+          />
+        </small>
+      {{/if}}
```
allthesignals: We could also make the DateDisplay component no-op if the date property is empty or invalid

PR #884, opened 2019-10-31
Better modal closing

andycochran:
Rahul's testing shows the modal is difficult to cancel. This PR makes it easier by allowing the user to click anywhere outside the content to close the modal.

### DIFF
--- a/app/styles/modules/_m-reveal-modal.scss
+++ b/app/styles/modules/_m-reveal-modal.scss
@@ -15,6 +15,14 @@
   }
 }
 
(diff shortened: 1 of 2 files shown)


andycochran:
> Interesting approach! So is it covering the entire screen with a div that sits above the page, but below the modal? I wonder if it's consistent across browsers.

@godfreyyeung Yup. It should be fine in any browser since the `div.reveal` has z-index which positions it above the new `div.reveal-overlay-target`.

PR #893, opened 2019-11-06
889: fix bug in single disposition option confirmation modal on hearing form

### LINKED ISSUE #889: Hearing details aren't populating on hearing confirmation modal

trbmcginnis:
Fix bug where, for assignments with only one disposition, data wasn't showing up in the confirmation modal. 

Closes #889

### DIFF
--- a/app/controllers/my-projects/assignment/hearing/add.js
+++ b/app/controllers/my-projects/assignment/hearing/add.js
@@ -86,7 +86,7 @@ export default class MyProjectsProjectHearingAddController extends Controller {
   @action
   async onConfirm() {
     const { dispositions } = this.model;
-    const allActions = this.get('allActions');
+    const allActions = this.get('allActions') || (dispositions.length <= 1);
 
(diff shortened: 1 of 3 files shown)


app/controllers/my-projects/assignment/hearing/add.js
```diff
@@ -86,7 +86,7 @@ export default class MyProjectsProjectHearingAddController extends Controller {
   @action
   async onConfirm() {
     const { dispositions } = this.model;
-    const allActions = this.get('allActions');
+    const allActions = this.get('allActions') || (dispositions.length <= 1);
```
allthesignals: Great! I think later on we'll want to centralize the `allActions` state into a computed property. That way we're not re-computing the `allActions` state in different places based on different criteria. Thanks for the quick turnaround!

trbmcginnis: Yes sounds like a good plan

PR #902, opened 2019-11-07
897 Use CRM Enum values for Recommendation values

godfreyyeung:
Refactor to use CRM Enum values for Recommendation values


app/components/recommendation-symbol.js
```diff
@@ -1,4 +1,20 @@
...
+  participantRecommendation = null;
+
+  @computed('participantRecommendation')
+  get icon() {
+    let icon = { style: '', color: 'light-gray' };
+    if ([717170000, 717170001].includes(this.participantRecommendation)) {
+      icon = { style: 'thumbs-up', color: 'green-muted' };
+    } else if ([717170002, 717170003].includes(this.participantRecommendation)) {
+      icon = { style: 'thumbs-down', color: 'red-muted' };
+    } else if ([717170002, 717170006, 717170008].includes(this.participantRecommendation)) {
```
allthesignals: Hardcoding these in worries me... let's figure out a better strategy for handling constants like this.

Issue #895, opened 2019-11-06
sub-milestones not showing up when submitting a hearing in staging


trbmcginnis:
sub-milestones are not showing up on tabs or on `show-project` page

trbmcginnis:
- sub-milestones not showing up was happening because `dcp_recommendationsubmittedbyname` field had not been added to the API
- Most of `dcp_recommendationsubmittedbyname` values in the `dcp_communityboarddisposition` table in the database are NULL, which breaks the `hearings-list-for-milestones-list` component because this field is necessary for matching dispositions to particular milestones
- shelving this for now until we find a better solution

Possible solutions:  
- Solution option: make sure that this field is filled out consistently in the future
- Solution option: we can choose a different field to use for matching. We do have `dcp_recommendationsubmittedby` field which is a contact id that matches the `contactid` field in the `contact` table. We can map `dcp_recommendationsubmittedby` to the user's contactid and then grab the corresponding `landUseParticipant` value. Create a new computed property in disposition model that does this. Looks like we're not currently grabbing the `contactid` from the API for the frontend `user` model so will have to add that as well

trbmcginnis:
Closing this because the refactor in #948 will fix it

trbmcginnis:
Closing this because the refactor in #948 will fix it

hannahkates:
This is still a problem. Hearings and recommendation submilestones aren't appears on the project profile.

trbmcginnis:
Original fix: use the field `fullname` from the contact table as a replacement for `dcp_recommendationsubmittedbyname`

Now we're encountering data issues:

In order for the hearings data to show up on the sub-milestones, `dcpPublichearinglocation` and `dcpDateofpublichearing` fields are required.

In order for recommendation data to show up on the sub-milestones, `dcpDateofvote`, `dcpVotinginfavorrecommendation`, `dcpVotingagainstrecommendation`, `dcpVotingabstainingonrecommendation`, and one of the recommendation fields (e.g. `dcpCommunityboarddisposition`) are required. 

It looks like even when the `statuscode` for a project is `1`, which means the recommendation has been submitted and each of these above required fields SHOULD be filled, the data shows that there are still a lot of NULL values.

trbmcginnis:
An Example project: https://zap-staging-lupp.planninglabs.nyc/projects/P2016Q0306

[...]

PR #906, opened 2019-11-08
895: check that dcpRecommendationsubmittedbyname is not NULL

trbmcginnis:
This is a temporary fix for Issue #895 

Because many of the values in the database for the field `dcpRecommendationsubmittedbyname` are `NULL`, `hearings-list-for-milestones-list` will often break when it tries to match this field with the corresponding milestone. 

This temporary fix is coupled with a backend change of adding `dcpRecommendationsubmittedbyname` as a field to dispositions in the API. 

New test written for the bug

To-Do: Figure out better way to fix this problem based on data we have

### DIFF
--- a/app/components/hearings-list-for-milestones-list.js
+++ b/app/components/hearings-list-for-milestones-list.js
@@ -68,13 +68,23 @@ export default class HearingsListForMilestonesListComponent extends Component {
     const dispositions = milestone.get('project.dispositions');
     const milestoneParticipantReviewLookup = this.get('milestoneParticipantReviewLookup');
 
+    const filteredDispositionsArray = [];
+
     // Iterate through ALL of the current project's dispositions.
     // Filter by IF a single disposition's dcpRecommendationsubmittedbyname matches the
     // current milestone's displayName based on the milestoneParticipantReviewLookup.
     // disposition.dcpRecommendationsubmittedbyname = e.g. 'QNBP'
     // disposition.dcpRecommendationsubmittedbyname.substring(2,4) = e.g. 'BP'
     // matching e.g. 'BP' with the milestoneParticipantReviewLookup provides 'Borough President Review'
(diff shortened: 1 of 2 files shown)


app/components/hearings-list-for-milestones-list.js
```diff
@@ -68,13 +68,23 @@ export default class HearingsListForMilestonesListComponent extends Component {
...
-    return dispositions.filter(disposition => milestoneParticipantReviewLookup[milestone.displayName] === disposition.dcpRecommendationsubmittedbyname.substring(2, 4));
+    dispositions.forEach(function(disposition) {
+      // make sure that dcpRecommendationsubmittedbyname is not NULL
+      if (disposition.dcpRecommendationsubmittedbyname) {
+        if (milestoneParticipantReviewLookup[milestone.displayName] === disposition.dcpRecommendationsubmittedbyname.substring(2, 4)) {
+          filteredDispositionsArray.push(disposition);
+        }
+      }
+    });
+    return filteredDispositionsArray;
```
allthesignals: You can use a .filter then .map but this is fine considering time constraints

tests/acceptance/hearings-list-for-milestones-list-shows-up-correctly-test.js
```diff
@@ -723,4 +723,103 @@ module('Acceptance | hearings list for milestones list shows up correctly', func
...
+        ],
+      }),
+    });
+
+    await visit('/my-projects/upcoming');
+
+    // make sure that milestones list on assignment card shows up
+    assert.ok(find('[data-test-milestone-id="12"]'), 'community board milestone');
+    assert.ok(find('[data-test-milestone-id="38"]'), 'borough board milestone');
+  });
```
allthesignals: Your tests have been a lifeline

app/components/hearings-list-for-milestones-list.js
trbmcginnis: @allthesignals Can you explain how that would fix the `NULL` error? The error was 'cannot find substring of null` so it would still error if I did that filter first.

PR #917, opened 2019-11-13
543 Document upload FE and tests

godfreyyeung:
- FE business logic for uploading files before submitting a recommendation. 
- Acceptance Tests along with mirage endpoint mocking

*TODO*:
- [x] end-to-end point and click tests w/ app connected to local ZAP-API after FE blockers are resolved.✅ thx @allthesignals !
(my-projects page currently down due to malformed assignments responses.)

- [ ] Keep an eye out for a case where a duplicate Sharepoint folder is created for a specific disposition instance

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ You will need the following things properly installed on your computer.
 ## Running / Development
 
 * `ember serve` - Uses dummy API responses provided by mirage
-* `ember serve --environment=devlocal` - Sends api calls to `localhost:3000`, assumes you are running [labs-zap-api](https://github.com/NYCPlanning/labs-zap-api) locally.
+* `HOST_API=http://localhost:3000 ember serve --environment=devlocal` - Sends api calls to `localhost:3000`, assumes you are running [labs-zap-api](https://github.com/NYCPlanning/labs-zap-api) locally.
 * Visit your app at [http://localhost:4200](http://localhost:4200).
 
 ### Create test scenarios
--- a/app/controllers/my-projects/assignment/recommendations/add.js
+++ b/app/controllers/my-projects/assignment/recommendations/add.js
@@ -4,6 +4,7 @@ import { alias } from '@ember/object/computed';
 import { inject as service } from '@ember/service';
 import lookupValidator from 'ember-changeset-validations';
 import Changeset from 'ember-changeset';
+import ENV from 'labs-zap-search/config/environment';
 import {
   bpDispositionForAllActionsValidations,
   cbDispositionForAllActionsValidations,
@@ -81,6 +82,9 @@ export default class MyProjectsProjectRecommendationsAddController extends Contr
   @service
   currentUser;
 
+  @service
+  fileQueue;
+
   @alias('model.dcpLupteammemberrole')
   participantType;
 
@@ -95,6 +99,15 @@ export default class MyProjectsProjectRecommendationsAddController extends Contr
   dispositionForAllActions = DispositionForAllActions.create();
 
   minDate = MINIMUM_VOTE_DATE;
+
+  queueName = 'recommendation';
+
+  submitError = false;
+
+  @computed('fileQueue', 'queueName')
+  get recommendationAddQueue() {
+    return this.fileQueue.find(this.queueName);
+  }
 
   @computed('dispositionForAllActions', 'participantType')
   get dispositionForAllActionsChangeset() {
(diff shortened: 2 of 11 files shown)


README.md
```diff
@@ -22,7 +22,7 @@ You will need the following things properly installed on your computer.
 ## Running / Development
 
 * `ember serve` - Uses dummy API responses provided by mirage
-* `ember serve --environment=devlocal` - Sends api calls to `localhost:3000`, assumes you are running [labs-zap-api](https://github.com/NYCPlanning/labs-zap-api) locally.
+* `HOST_API=http://localhost:3000 ember serve --environment=devlocal` - Sends api calls to `localhost:3000`, assumes you are running [labs-zap-api](https://github.com/NYCPlanning/labs-zap-api) locally.
```
allthesignals: We could simplify this a little more by having another environment variable of our own:

`HOST_API=http://localhost:3000 MIRAGE_ENABLED=false ember serve`

Take a look at how this one does: https://github.com/adlius/ember-osf-web/blob/develop/config/environment.js#L306-L308

I can set this up if you prefer, let me know

app/controllers/my-projects/assignment/recommendations/add.js
```diff
@@ -257,49 +270,83 @@ export default class MyProjectsProjectRecommendationsAddController extends Contr
...
+        // next call of upload() on that file.
+        const fileUploadResponses = await Promise.all(fileUploadPromises); // eslint-disable-line
+
+        const filesUploadedToDispo = fileUploadResponses.every(res => res.status === 200);
 
-    this.dispositions.forEach((disposition) => {
-      if (this.allActions) {
-        disposition.set(targetField, this.dispositionForAllActions.recommendation);
+        uploadResults.push(filesUploadedToDispo);
+      }
```
allthesignals: I really don't think we should use for loops. The reason is that turns everything synchronous. So each await that gets triggered in the loop waits for the last one. See this: https://zellwk.com/blog/async-await-in-loops/

This can be achieved with array maps, reducers, etc.

app/controllers/my-projects/assignment/recommendations/add.js
```diff
@@ -257,49 +270,83 @@ export default class MyProjectsProjectRecommendationsAddController extends Contr
...
+      this.dispositions.forEach((disposition) => {
+        if (this.allActions) {
+          disposition.set(targetField, this.dispositionForAllActions.recommendation);
+          disposition.setProperties({
+            dcpVotinginfavorrecommendation: this.dispositionForAllActions.dcpVotinginfavorrecommendation,
+            dcpVotingagainstrecommendation: this.dispositionForAllActions.dcpVotingagainstrecommendation,
+            dcpVotingabstainingonrecommendation: this.dispositionForAllActions.dcpVotingabstainingonrecommendation,
+            dcpTotalmembersappointedtotheboard: this.dispositionForAllActions.dcpTotalmembersappointedtotheboard,
+            dcpConsideration: this.dispositionForAllActions.dcpConsideration,
+          });
```
allthesignals: You can make this a little easier to grok by using destructuring and object assignment shorthand:

const { a,b,c,d } = this.dispositionForAllActions;

disposition.setProperties({
a,b,c,d
});

tests/unit/controllers/my-projects/assignment/recommendations/add-test.js
```diff
@@ -154,6 +154,10 @@ module('Unit | Controller | my-projects/assignment/recommendations/add', functio
 
     controller.transitionToRoute = function() { return true; };
 
+    controller.set('recommendationAddQueue', {
+      files: [],
+    });
+
```
allthesignals: I'd avoid these unit tests but this is fine

README.md
godfreyyeung: Thank you! I'll put it in the backlog, and see if I can pick it up later

app/controllers/my-projects/assignment/recommendations/add.js
godfreyyeung: Great article, thank you for the share 🙇

allthesignals: Alright, but I would say that this is still possible _without_ a for loop, and that if you needed to control the concurrency, you can do that through other more expressive means. We can still merge this but I'm personally not convinced there's no way to avoid a synchronous for-loop.

allthesignals:
Sorry going ahead and merging this

PR #932, opened 2019-11-14
Create disposition specific queues with copies of each file

godfreyyeung:
This PR breaks out the universal file queue into queues for each disposition. 
 
Each disposition queue will have a copy of each file uploaded. This is necessary to prevent the file from being flushed from every one of its parent queues (and previously from the universal queue) after it is successfully uploaded once for the first disposition.

This may also help with the any future plans to upload different files to each disposition.

### DIFF
--- a/app/controllers/my-projects/assignment/recommendations/add.js
+++ b/app/controllers/my-projects/assignment/recommendations/add.js
@@ -5,6 +5,7 @@ import { inject as service } from '@ember/service';
 import lookupValidator from 'ember-changeset-validations';
 import Changeset from 'ember-changeset';
 import ENV from 'labs-zap-search/config/environment';
+import File from 'ember-file-upload/file';
 import {
   bpDispositionForAllActionsValidations,
   cbDispositionForAllActionsValidations,
@@ -100,13 +101,19 @@ export default class MyProjectsProjectRecommendationsAddController extends Contr
 
   minDate = MINIMUM_VOTE_DATE;
 
+  // currently only used to name the universal queue
   queueName = 'recommendation';
 
   submitError = false;
 
-  @computed('fileQueue', 'queueName')
-  get recommendationAddQueue() {
-    return this.fileQueue.find(this.queueName);
+  // Returns an object with an entry for each disposition and its corresponding file queue.
+  // Keys are the disposition id, values are the disposition's file queue.
+  @computed('fileQueue', 'dispositions')
+  get queuesByDisposition() {
+    return this.dispositions.reduce((queuesByDisposition, disposition) => {
+      queuesByDisposition[disposition.id] = this.fileQueue.create(disposition.id);
+      return queuesByDisposition;
+    }, {});
   }
 
   @computed('dispositionForAllActions', 'participantType')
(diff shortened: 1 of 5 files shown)


app/controllers/my-projects/assignment/recommendations/add.js
```diff
@@ -100,13 +101,18 @@ export default class MyProjectsProjectRecommendationsAddController extends Contr
...
-  @computed('fileQueue', 'queueName')
-  get recommendationAddQueue() {
-    return this.fileQueue.find(this.queueName);
+  @computed('fileQueue', 'dispositions')
+  get queuesByDisposition() {
+    const queuesByDisposition = {};
+    this.dispositions.forEach((disposition) => {
+      queuesByDisposition[disposition.id] = this.fileQueue.create(disposition.id);
+    });
+    return queuesByDisposition;
```
allthesignals: You could alternatively return this as a `.reduce` off of `this.dispositions`

app/controllers/my-projects/assignment/recommendations/add.js
```diff
@@ -263,6 +269,16 @@ export default class MyProjectsProjectRecommendationsAddController extends Contr
...
   }
 
+  // adds the uploaded file to each disposition queue
+  @action
+  async addFileToDispositionQueues(file) {
+    this.dispositions.map(async (disposition) => {
+      // use this private/undocument .blob property to duplicate the file
+      const newFile = await File.fromBlob(file.blob);
+      this.queuesByDisposition[disposition.id].push(newFile);
+    });
```
allthesignals: In this case, `map` is for iterating over an array and _returning a new array_ based on the return value. I think it's unclear here when you use map because you're not using the return value, and your callback function implicitly returns a promise (though it doesn't explicitly return something inside).

app/controllers/my-projects/assignment/recommendations/add.js
```diff
@@ -274,11 +290,10 @@ export default class MyProjectsProjectRecommendationsAddController extends Contr
     const uploadResults = [];
 
     try {
-      // loop through dispositions, uploading all files to each one
       for (let i = 0; i < this.dispositions.length; i += 1) {
         const disposition = this.dispositions.objectAt(i);
 
-        const fileUploadPromises = this.recommendationAddQueue.files.map(file => file.upload(`${ENV.host}/document`, {
+        const fileUploadPromises = this.queuesByDisposition[disposition.id].files.map(file => file.upload(`${ENV.host}/document`, {
```
allthesignals: Can there ever be a dispo ID that doesn't have a queue

app/controllers/my-projects/assignment/recommendations/add.js
godfreyyeung: Pro suggestion, thanks!

app/controllers/my-projects/assignment/recommendations/add.js
godfreyyeung: Hmm yeah you're right. I'm not sure what I would return from the action though, since nothing would ever use the action's return value. Maybe I'll at least make the map's return of a Promise more explicit, though.

app/controllers/my-projects/assignment/recommendations/add.js
godfreyyeung: I don't think so, each disposition attached to the assignment should have a file queue I guess. 

@hannahkates would there ever be a disposition which shouldn't have files uploaded to it?

app/controllers/my-projects/assignment/recommendations/add.js
godfreyyeung: updated!

app/controllers/my-projects/assignment/recommendations/add.js
godfreyyeung: @allthesignals updated, take a look if you get a free moment!

PR #941, opened 2019-11-14
add message to all empty tabs

andycochran:
This PR makes sure all the tabs have an message to provide the user some feedback/action if there are no projects in the list.


andycochran:
@hannahkates Updated.

PR #944, opened 2019-11-15
900: fix waive hearings by adding .save()

### LINKED ISSUE #900: Can't opt out of hearings on the Upcoming tab
If you go to the Upcoming tab and click "Opt Out" then confirm "Opt Out" again in the modal, the modal closes, but nothing changes in the project UI — you still see the post/opt-out buttons.

trbmcginnis:
- update the function that sets hearings to waived (`dcpIspublichearingrequired` = `No`) to include `.save()` instead of just setting the new value
- update tests to check for error handling
- move `waive-hearings-popup` to outside of the if statements on the project cards so that the component is not torn down before a user has submitted their info
- when tests forced an error, for some reason the `disposition.dcpIspublichearingrequired` values were still getting set to `No`, so the error function also resets this value so that the four computeds that check for hearings waived/submitted on the assignments model are not incorrect

Am open to fixes to this! Might require a bigger refactor in the future.

Closes #900

### DIFF
--- a/tests/acceptance/user-can-waive-hearings-test.js
+++ b/tests/acceptance/user-can-waive-hearings-test.js
@@ -25,6 +25,10 @@ module('Acceptance | user can waive hearings', function(hooks) {
     await invalidateSession();
   });
 
+  hooks.beforeEach(async function() {
+    await authenticateSession();
+  });
+
   test('user can waive hearings on to-review page', async function(assert) {
     this.server.create('assignment', {
       id: 4,
@@ -49,8 +53,6 @@ module('Acceptance | user can waive hearings', function(hooks) {
       }),
     });
 
-    await authenticateSession();
-
     await visit('/my-projects/to-review');
 
     assert.ok('[data-test-button="submitHearing"]');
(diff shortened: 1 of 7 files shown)


tests/acceptance/user-can-waive-hearings-test.js
```diff
@@ -70,6 +72,8 @@ module('Acceptance | user can waive hearings', function(hooks) {
     assert.ok(find('[data-test-hearings-waived-message]'));
     assert.ok(find('[data-test-button="submitRecommendation"]'));
 
+    assert.equal(this.server.db._collections[2]._records[0].dcpIspublichearingrequired, 'No');
```
allthesignals: Let's look into this more, there _should_ be a public getter

allthesignals: @trbmcginnis 
Screen Shot 2019-11-15 at 2 14 29 PM

I'm not sure why you were unable to use the public method for this. I can't recreate this. One think that may have happened is that you were console logging `this.server.db`, and didn't see a concrete property. I think these collections are defined as getters, so you won't see them except at the bottom faded out.

allthesignals: I'm going to fix this in an upcoming PR. Thanks!

PR #951, opened 2019-11-15
Configure which search field to use

### LINKED ISSUE #936: recommendation power-select filtering errors
Filtering on the recommendation power-select is erroring, complaining about needing a searchField value

allthesignals:
Closes #936


allthesignals:
@godfreyyeung will do! thanks

Issue #948, opened 2019-11-15
Refactor sub-milestones


trbmcginnis:
Do a LEFT JOIN in the backend `dcp_communityboarddisposition.dcp_recommendationsubmittedby` and `contact.fullname` so that we have access to the fullname field on each disposition. This field will take the place of `dcp_recommendationsubmittedbyname` since many of the rows in this field are NULL. 

~Show list of hearings for a project. If participant type information is available, display it. ~At the ~moment we cannot link hearing dispos to milestones, so we will show them separately. ~@andycochran please advise a design workaround for this...~ @hannahkates found a way to do this!~

~One missing piece is `dcpRecommendationsubmittedbyname`. Is it being filled in by the hearing submission form? We should make sure this happens. ~

~That said, we think we should still do these things in the future, data quality permitting.~

~Enhancements:~

- ~No longer use disposition model field `dcpRecommendationsubmittedbyname` because a lot of its values are NULL~
- ~Create a model relationship between milestones and dispositions? One milestone will have many dispositions~

hannahkates:
Hmm @trbmcginnis thinking about this more, I'm not sure if it'll work. For the submilestones, we want ALL the submilestones for all LUPs on a given project. The computer value on the assignment model is dependent on the specific user that's logged in. There may be another way around this, but I don't think we can use assignment.

trbmcginnis:
@hannahkates Yeah realized this too :/, we're thinking about doing a model relationship between milestones and dispositions

trbmcginnis:
SELECT disp.dcp_recommendationsubmittedby, contact.fullname FROM dcp_communityboarddisposition AS disp
LEFT JOIN contact ON dcp_recommendationsubmittedby = contact.contactid

PR #955, opened 2019-11-18
948: refactor sub-milestones to use fullname field

trbmcginnis:
- Refactor sub-milestones to use `fullname` field (which is originally sourced from a left join with the contact table in the API)
- `fullname` takes the place of `dcpRecommendationsubmittedbyname`, which is oftentimes NULL in the database
- `fullname` is used to associate a specific disposition with a milestone AND to display which user submitted a hearing/vote
- there are already tests that make sure that nothing breaks when `fullname` is NULL in the database

Addresses #948


app/components/hearings-list-for-milestones-list.js
```diff
@@ -68,23 +68,18 @@ export default class HearingsListForMilestonesListComponent extends Component {
...
-          filteredDispositionsArray.push(disposition);
-        }
-      }
+    const currentMilestoneDispositions = dispositions.filter(function(disposition) {
+      // make sure we aren't grabbing any null values; null values break substring
+      const participantType = disposition.fullname != null ? disposition.fullname.substring(3, 5) : '';
+      return milestoneParticipantReviewLookup[milestone.displayName] === participantType;
     });
-    return filteredDispositionsArray;
+    return currentMilestoneDispositions;
```
allthesignals: _great_ — you can even go a step further and just return the `dispositions.filter`.

app/components/hearings-list-for-milestones-list.js
```diff
@@ -111,16 +106,15 @@ export default class HearingsListForMilestonesListComponent extends Component {
     // This participantRecommendationType is passed into `deduped-votes-list` to assure that we are deduplicating
     // by the correct recommenation field.
     const milestoneParticipants = currentMilestoneDispositions.map(disp => ({
-      landUseParticipantFullName: disp.recommendationSubmittedByFullName,
+      landUseParticipantFullName: disp.fullNameLongFormat,
       participantRecommendationType: partRecType, // e.g. 'dcpCommunityboardrecommendation'
-      landUseParticipantType: disp.dcpRecommendationsubmittedbyname.substring(2, 4), // e.g. 'CB'
+      landUseParticipantType: disp.fullname.substring(3, 5), // e.g. 'CB'
```
allthesignals: might wanna comment this logic more

trbmcginnis: will do!

app/components/hearings-list-for-milestones-list.js
trbmcginnis: will do :)

PR #959, opened 2019-11-18
Computed attrs for state and status codes

allthesignals:
This PR adds some "computed attributes" for posting into the system. 

kudos to @spencer516 for the mere _7 lines_ of code it took to get a computed attr working.

### DIFF
--- a/tests/unit/models/disposition-test.js
+++ b/tests/unit/models/disposition-test.js
@@ -1,6 +1,6 @@
 import { module, test } from 'qunit';
 import { setupTest } from 'ember-qunit';
-import { STATUSCODES } from 'labs-zap-search/models/disposition';
+import { STATUSCODES, STATECODES } from 'labs-zap-search/models/disposition';
 import { RECOMMENDATION_OPTIONSET_BY_PARTICIPANT_TYPE_LOOKUP } from 'labs-zap-search/controllers/my-projects/assignment/recommendations/add';
 
 module('Unit | Model | disposition', function(hooks) {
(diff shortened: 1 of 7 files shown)


tests/unit/models/disposition-test.js
```diff
@@ -1,13 +1,50 @@
...
   // Replace this with your real tests.
-  test('it exists', function(assert) {
+  test('it exists', async function(assert) {
     const store = this.owner.lookup('service:store');
     const model = store.createRecord('disposition', {});
+
     assert.ok(model);
   });
+
+  test('when participant submits/opts out, statuscode is "saved"', async function(assert) {
```
godfreyyeung: Maybe the title should be when participant opts out/submits, statuscode becomes "saved"/"submitted"

tests/unit/models/disposition-test.js
```diff
@@ -1,13 +1,50 @@
...
+    assert.equal(model.statuscode, STATUSCODES.findBy('Label', 'Submitted').Value);
+  });
+
+  test('when statuscode is draft or saved, statecode is "active"', async function(assert) {
+    const store = this.owner.lookup('service:store');
+    const model = store.createRecord('disposition', {
+      statuscode: STATUSCODES.findBy('Label', 'Draft').Value,
+    });
+
+    assert.equal(model.statecode, STATECODES.findBy('Label', 'Inactive').Value);
```
godfreyyeung: Weird, shouldn't statecode be Active if statuscode is Daft? The `statecode()` computed looks right, though...

tests/unit/models/disposition-test.js
allthesignals: good catch!

tests/unit/models/disposition-test.js
allthesignals: yup!

PR #995, opened 2019-11-22
only show rec submission dates if they exist

andycochran:
On the Reviewed and Archive tabs, we should only show the recommendation submission dates if they exist.


andycochran:
@allthesignals I did an integration test. Is that cool?

PR #1004, opened 2019-11-22
962 Use assignment specific queue (patch files carrying across assignments)

godfreyyeung:
The recommendation form will now rely on an "assignment" queue for tracking files prior to the user clicking "submit". Only upon submit will the files be copied over to disposition-specific queues.

Upon submit, the assignment-specific queue is also flushed, so that files will no longer be carried across assignments.

This will also help support file removal (since the UI is now tied to one main queue), and eventually action-specific file uploads. 

Todos:
   - The submitRecommendation action is getting a bit gnarly. Investigate breaking it down.


app/controllers/my-projects/assignment/recommendations/add.js
```diff
@@ -101,19 +101,38 @@ export default class MyProjectsProjectRecommendationsAddController extends Contr
...
+  // disposition queues.
+  // This queue is also used to list files next to the "upload" button.
+  // TODOS:
+  //    - Feature to remove files from this queue.
+  //    - When implementing "by action" file upload, only copy files from this queue
+  //      to dispositions queues if allActions === true.
+  @computed('fileQueue', 'assignmentQueueName')
+  get assignmentQueue() {
+    return this.fileQueue.find(this.assignmentQueueName) || this.fileQueue.create(this.assignmentQueueName);
+  }
```
allthesignals: Nice splitting things out a bit more... I am worried about this though, because it breaks the declarative nature of computed properties through the use of that _conditionally_-invoked `create` method. A computed property that mutates some other data elsewhere introduces side-effects, makes stuff hard to debug. I'm gonna see if I can find another approach.. Thanks!

allthesignals:
I've gone ahead and merged develoop

PR #1006, opened 2019-11-22
add use-a-better-browser message for IE

andycochran:
This PR adds the method used in https://github.com/NYCPlanning/labs-applicantmaps/commit/97d47555b73f785f2cf51b4e6fe29250be6cbb18 to provide a "use a modern browser" message. 

**Note: This will also block IE11, not just IE10.** 

If we're to support IE 11, we'll need to use a different method.


andycochran:
@allthesignals I realized that this addon has a service property for looking at the specific user agent info. So I've updated it to check the major version with `(lte (user-agent "browser.info.major") 10)` in addition to `(user-agent "browser.isIE")`. IE11, will now work as intended.

PR #1024, opened 2019-11-26
1003: use new fields for milestone duration and time remaining

### LINKED ISSUE #1003: Duration and days remaining should not be calculated
Our app is computing “**x** of **y** days remain” with `.diff(moment(), 'days')` — e.g.: 

```
  @computed('assignment.toReviewMilestonePlannedCompletionDate')
  get timeRemaining() {
    return moment(this.assignment.toReviewMilestonePlannedCompletionDate).diff(moment(), 'days');
  }
```

Unfortunately, this wrong if the review period spans a holiday. CRM has a value for a milestone's duration and logic tat computes a value for days remaining which avoids including holidays. We should pull these values from CRM instead of computing. 

The `dcp_projectmilestones ` table includes two "remaining days" — one of which I assume we should be using for `timeRemaining`: 
- ❌`dcp_remainingplanneddays` 
- ✅`dcp_remainingplanneddayscalculated`

The `dcp_projectmilestones ` table includes a bunch for "duration" ~— not sure which of these we should use~: 
- ❌`dcp_legalduration`
- ❌`dcp_warningduration`
- ❌`dcp_warningduration`
- ❌`dcp_actualduration`
- ✅`dcp_actualdurationasoftoday` <— confirmed below
- ❌`dcp_goalduration`
- ❌`dcp_goalduration2`
- ❌`dcp_goalduration3`

@gopal2020, @Eabrante, @anidamurati, can you please confirm which of these fields we should use?

trbmcginnis:
- add new fields `dcpRemainingplanneddayscalculated` and `dcpActualdurationasoftoday`
- update `to-review` and `reviewed` tabs to use the new these new fields
- update display logic in case one of these fields is NULL
- add new acceptance tests

Closes #1003 with backend changes PR https://github.com/NYCPlanning/zap-api/pull/55

### DIFF
--- a/app/models/assignment.js
+++ b/app/models/assignment.js
@@ -153,6 +153,20 @@ export default class AssignmentModel extends Model {
     return dcpPlannedcompletiondate;
   }
 
+  @computed('tab', 'dcpLupteammemberrole', 'project.milestones')
+  get toReviewMilestoneTimeRemaining() {
+    const participantMilestoneId = this.milestoneConstants.referralIdentifierByAcronymLookup[this.dcpLupteammemberrole];
+    const { dcpRemainingplanneddayscalculated } = this.project.get('milestones').find(milestone => milestone.dcpMilestone === participantMilestoneId) || {};
+    return dcpRemainingplanneddayscalculated;
+  }
+
+  @computed('tab', 'dcpLupteammemberrole', 'project.milestones')
+  get toReviewMilestoneTimeDuration() {
+    const participantMilestoneId = this.milestoneConstants.referralIdentifierByAcronymLookup[this.dcpLupteammemberrole];
+    const { dcpActualdurationasoftoday } = this.project.get('milestones').find(milestone => milestone.dcpMilestone === participantMilestoneId) || {};
+    return dcpActualdurationasoftoday;
+  }
+
   // If `tab` is 'reviewed'...
   //   - these start/end dates come from the current In Progress milestone
   //   - an array of milestone dates is returned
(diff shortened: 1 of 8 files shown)


app/models/assignment.js
```diff
@@ -153,6 +153,20 @@ export default class AssignmentModel extends Model {
...
+    const { dcpRemainingplanneddayscalculated } = this.project.get('milestones').find(milestone => milestone.dcpMilestone === participantMilestoneId) || {};
+    return dcpRemainingplanneddayscalculated;
+  }
+
+  @computed('tab', 'dcpLupteammemberrole', 'project.milestones')
+  get toReviewMilestoneTimeDuration() {
+    const participantMilestoneId = this.milestoneConstants.referralIdentifierByAcronymLookup[this.dcpLupteammemberrole];
+    const { dcpActualdurationasoftoday } = this.project.get('milestones').find(milestone => milestone.dcpMilestone === participantMilestoneId) || {};
+    return dcpActualdurationasoftoday;
+  }
```
allthesignals: Makes sense, good choice. This computed depends on properties from assignment and project.

app/templates/components/to-review-project-card.hbs
```diff
@@ -13,16 +13,22 @@
     </div>
     <div class="cell medium-4 large-3 xlarge-2">
       <div class="text-center medium-margin-right medium-margin-left">
-        <p class="tiny-margin-bottom">{{participant-type-label assignment.dcpLupteammemberrole}} Review</p>
+        <p class="tiny-margin-bottom" data-test-milestone-label="{{assignment.id}}">{{participant-type-label assignment.dcpLupteammemberrole}} Review</p>
```
allthesignals: good... but is the label really the assignment id?

Issue #1017, opened 2019-11-25
Stage: When entering out of bounds numbers for the vote fields, project is not submitting and is there is a continuous loop/buffering 


ksinghprod:
Browser; Chrome

Expected result: If there are fields that contain out of bound numbers, there should be an error message indicating which fields they are prior to submission of a recommendation

Actual Result: When entering out of bounds numbers for the vote fields, project is not submitting and is there is a continuous loop/buffering 

(I send a screen recording to gopal on this issue) 

![image](https://user-images.githubusercontent.com/57362229/69569462-b46af280-0f8b-11ea-8466-7e791774ed1b.png)

allthesignals:
Need a validation

allthesignals:
Addressed by PR #1027 — front end validates the field.

hannahkates:
This is resolved, QA team please check.

ksinghprod:
@andycochran @hannahkates 

same issue as #973 

![image](https://user-images.githubusercontent.com/57362229/69997939-47270680-1523-11ea-8da8-fef9e6ef27ea.png)

ksinghprod:
Matt here:

The issue here is with error handling. After an error, it should allow subsequent tries.

allthesignals:
PR up

allthesignals:
@ksinghprod please retest in about 10-15 minutes from now.

gopal2020:
This bug is fixed
![image](https://user-images.githubusercontent.com/57265341/70178994-36a19800-16ab-11ea-9873-e3eeb726abbc.png)

PR #1039, opened 2019-12-03
Do not show hearing button if there are no associated dispositions

allthesignals:
Addresses #1017 by preventing display of the hearing button.

The assignment in the issue did not have any associated dispositions. This might be a data or querying issue.


app/templates/components/to-review-project-card.hbs
```diff
@@ -43,37 +43,38 @@
...
-          <div class="cell medium-shrink">
-            <button
-              class="button expanded gray"
-              onClick={{action "openOptOutHearingPopup"}}
-              data-test-button="optOutHearingOpenPopup"
-            >
-              {{fa-icon 'calendar-times' size="lg" transform="shrink-5 down-1"}}
-              <small>Opt Out</small>
-            </button>
+      {{#if this.assignment.dispositionsByRole}}
```
andycochran: Instead of not showing the hearings button at all, let's make it disabled (like how the recommendation button is conditionally disabled).

allthesignals: @andycochran sure. this will need to be applied to the upcoming tab though.

app/models/assignment.js
```diff
@@ -41,9 +41,9 @@ export default class AssignmentModel extends Model {
 
   // if the each dcpPublichearinglocation and dcpDateofpublichearing properties are filled in dispositions array,
   // then hearings have been submitted for that project
-  @computed('dispositions.@each.{dcpPublichearinglocation,dcpDateofpublichearing}')
+  @computed('dispositionsByRole.@each.{dcpPublichearinglocation,dcpDateofpublichearing}')
   get hearingsSubmitted() {
-    const dispositions = this.get('dispositions');
+    const dispositions = this.get('dispositionsByRole');
```
godfreyyeung: I know this was from a previous PR, but I'm confused by why we need the `dispositionsByRole` computed. Shouldn't the zap-api already deliver an assignments with its attached dispositions filtered to `assignment.dcp_Lupteammemberrole`?

allthesignals: @godfreyyeung it should but it's not and I don't know why. I think @hannahkates was looking at it one point. I want to make this explicit check so we're not submitting dispositions against the wrong roles.

app/templates/components/to-review-project-card.hbs
allthesignals: I'm going to open this in another PR this needs to get merged now.

app/models/assignment.js
allthesignals: There's also no actual explicit join table between a lup team and disposition/contact. It's inferred through these fields.

PR #1058, opened 2019-12-06
1057 allow retry

### LINKED ISSUE #1020: Stage: LUP User is unable to submit a hearing with multiple hearing
Browser: Chrome

Expected Result: LUP user should be able to submit a public hearing consisting of multiple hearings, when selecting submitting each one manually 

Actual Result: LUP User is getting an "Oops something went wrong, can you try again?" message

Seems like the app is in crash state after the error response 

(This time in entered everything correctly)

image

allthesignals:
Handle errors and allow retries on error responses from server.

This likely duplicates some documents in the upload process but nevertheless allows upload retries.

Closes #1020, but we should refactor as soon as humanly possible.


allthesignals:
@gopal2020 @ksinghprod pushing this patch to staging

Issue #1048, opened 2019-12-04
Disposition data isn't showing up on public profiles


hannahkates:
Now that we've promoted develop to master, real projects in production that have real dispositions filled in should be showing the disposition hearing and rec details on the project profile. They're not showing up.

Example: 
https://zap.planning.nyc.gov/projects/P2016Q0306
https://nycdcppfs.crm9.dynamics.com/main.aspx#850422025

#1048, #1045, #987 should all be fixed in coordination.

![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/6e5fe2a1-4310-4616-aab6-e519c76772ea)

trbmcginnis:
Looks like this is being caused by an error with `hearingsSubmitted` on `hearings-list-for-milestones-list` component, which is coming back false when it should be true

allthesignals:
I'll take a look

allthesignals:
It looks like you already have a passing test for this in your branch :O

Issue #987, opened 2019-11-22
Thumbs up or down does not capture intricacies of CB recommendation


hannahkates:
Instead of only showing a red thumbs down or green thumbs up, we should find a way to communicate more ambivalence about a project if the project is approved/disapproved with modifications. Suggestions have included using the color yellow instead of red or green or a rotated thumb.

#1048, #1045, #987 should all be fixed in coordination.

hannahkates:
Examples of styling challenges:
- BP reviews shouldn't have "vote" in labels; they don't vote.
![image](https://user-images.githubusercontent.com/13967925/70574508-43723000-1b72-11ea-996c-ad54ba8ed1e4.png)

PR #1068, opened 2019-12-09
1048: show sub-milestones on a disposition-by-disposition basis

trbmcginnis:
**NOTE: This shouldn't be promoted into master until #987 is also implemented.**

### Changes
While users that submit hearings and recommendations through our app are not allowed to submit information for only SOME dispositions but instead are forced to submit information for ALL dispositions at the same time, we originally displayed sub-milestones assuming that all dispositions for one user on one project would have filled fields. This is not the case with the CRM data so this PR updates the sub-milestones to check for the existence of sub-milestones on a disposition-by-disposition basis instead.

*Each hearing sub-milestone should have a date and a location*

*Each vote sub-milestone should have a date, votes in favor, votes against, votes abstaining, and a participant recommendation*

- `hearingsSubmitted` and `votesSubmitted` on `hearings-list-for-milestones-list.js` (NOTE: these are different from the properties of the same name on the assignment model) have been updated to check that at least ONE disposition within a user's dispositions for that project has the required fields filled. These properties are used to display the ENTIRE sub-milestone, including the title of the sub-milestone
- `dispHasAllFields` is a new property on the `newDispositionsArray` in both `deduped-votes-list.js` and `deduped-hearings-list.js`. These check that the current disposition has all required fields, and then displays or does not display each disposition based on `dispHasAllFields` being `true` or `false` 
- Because checking that date objects are truthy can be tricky (an invalid date object might still return truthy), we must convert dates to strings before checking. `toString()` will throw an error if the date field is null, so we must first assure that the date field is not null. There are updated checks for this throughout the hearings and votes list components

### Tests
Acceptance tests were added to check that when any of these fields are null or an empty string (two fields for hearings, and five fields for votes) that the sub-milestones will display as intended and that no error will occur

Addresses #1048


hannahkates:
I'm running the front-end locally with the staging API to check that this: 
- Hearings are showing up! Yay!
- Recommendations aren't though

Peninsula Hospital Redevelopment Plan (P2016Q0306) has real disposition data, including submitted recommendations, but the recs are missing on the project profile timeline.
image

Issue #874, opened 2019-10-31
Re-enable the "Actions" filter; debug issues


hannahkates:
We hid the filter in a previous PR.

We need to turn it back on and see if it works. If it doesn't, debug and see why.

Here's where the filter was turned off: https://github.com/NYCPlanning/labs-zap-search/blob/develop/app/templates/show-geography.hbs#L247-L272

Some notes on observed behavior:

> On prod, if you apply the action filter for "EC - Enclosed Sidewalk Cafes", you get ~60 results. If you do the same on staging, you get zero results. Same for other action types.

hannahkates:
Once the API fix is implemented, we need to unhide the action filter in the front end.

hannahkates:
API changed were implemented in https://github.com/NYCPlanning/zap-api/pull/57. Now we need to make the front end changes.

PR #1101, opened 2020-02-18
874 fix action filter

trbmcginnis:
### Changes:
Previously the action filter on the projects list page was not working and had to be removed. The commit where it was removed: d95b8e7.

This PR re-adds the action filter, which is now working again. The reason for the filter break is not known.

### Tests:
This PR adds a test in `filter-checkbox-test` acceptance test to assure that query parameters are changing appropriately when a user interacts with the action filter. This test does not actually test the full filter functionality, since that is handled in the backend. The actual filter functionality was tested manually.

Addresses #874

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,10 @@ The frontend will then send api calls to `localhost:3000`, assuming you are runn
 
 Sign in by going to `localhost:4200/login#access_token=test`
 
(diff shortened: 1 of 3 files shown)


allthesignals:
https://www.youtube.com/watch?v=PIsZQEhDFz8

PR #1104, opened 2020-02-18
1103: add heroku layers api passthrough to mirage config to fix tests

trbmcginnis:
Add heroku layers api passthrough to mirage config to fix tests

### DIFF
--- a/mirage/config.js
+++ b/mirage/config.js
@@ -11,6 +11,7 @@ export default function () {
   this.passthrough('https://raw.githubusercontent.com/**');
   this.passthrough('https://tiles.planninglabs.nyc/**');
   this.passthrough('https://layers-api-staging.planninglabs.nyc/**');
+  this.passthrough('https://labs-layers-api-staging.herokuapp.com/v1/base/style.json');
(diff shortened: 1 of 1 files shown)


mirage/config.js
```diff
@@ -11,6 +11,7 @@ export default function () {
   this.passthrough('https://raw.githubusercontent.com/**');
   this.passthrough('https://tiles.planninglabs.nyc/**');
   this.passthrough('https://layers-api-staging.planninglabs.nyc/**');
+  this.passthrough('https://labs-layers-api-staging.herokuapp.com/v1/base/style.json');
```
godfreyyeung: What's the difference between `https://layers-api.planninglabs.nyc/v1/base/style.json` and `https://labs-layers-api-staging.herokuapp.com/v1/base/style.json`? I noticed CEQR uses the first one.

trbmcginnis: Not sure, I didn't make the original changes that affected this, but having both of them there is necessary to get tests to pass

godfreyyeung: Ok! 👌

Issue #925, opened 2019-11-13
Update build-url helper to use online ZR


hannahkates:
We use a helper to generate the URLs provided for zoning actions on the project profiles. The helper assumes we should send the user to DCP's website. We should update the helper to use the new online ZR instead. However, we need more info about the online ZR URL structures and linking before we can do this

https://github.com/NYCPlanning/labs-zap-search/blob/develop/app/helpers/build-url.js

Screenshot of where links are provided on project profile:
![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/3e4e76d3-4d19-4135-900b-9fee2d8af6c4)

andycochran:
Here's an example of deep-linking to sections of the online Zoning Resolution site: 

https://zr.planning.nyc.gov/article-iii/chapter-3#33-012

PR #1105, opened 2020-02-18
925: Update zoning resolution URLs to use Online ZR

bfreeds:
Zoning Resolution links to DCPs website are now broken.
image. 

This commit directs users to the new Online ZR website instead, by
bringing them to the search page if the "DCP zoningresolution" value is
a zoning resolution (ex. "74-533").  

image

Because "DCP zoningresolution" value can be "AppendixA, AppendixB, etc."
this commit hardcodes the Appendix URLs, which is a better UX
than getting sent to a search page with no results.

This is a quick fix for broken links while we figure out a better way
to directly link users to the exact Zoning Resolution.

Addresses #925

### DIFF
--- a/app/helpers/build-url.js
+++ b/app/helpers/build-url.js
@@ -1,10 +1,5 @@
 import { helper } from '@ember/component/helper';
 
-function pad(string, size) {
-  while (string.length < (size || 2)) { string = `0${string}`; }
-  return string;
-}
-
 function bblDemux(bbl) {
   const boro = bbl.substring(0, 1);
   const block = parseInt(bbl.substring(1, 6), 10);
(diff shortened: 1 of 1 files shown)


app/helpers/build-url.js
```diff
@@ -1,10 +1,5 @@
 import { helper } from '@ember/component/helper';
 
-function pad(string, size) {
```
bfreeds: Removed this function, as it was not used anywhere.

bfreeds:
@allthesignals @hannahkates does it make sense for me to merge this into `develop`, or should this be treated as a ["hot fix"](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) and brought directly to `master`?
image

allthesignals:
@bfreeds how would this affect later merge of develop into master?

Issue #1019, opened 2019-11-25
Stage: When Selecting a community board recommendation outcome, fields are getting validated when selecting the disapproved or Received after clock Expired option


ksinghprod:
Browser: Chrome

Expected Result: When LUP user selects disapproved or Received after clock expired nothing should happen. Only the selected choice should appear (vote fields should not get validated)

Actual Result: When Selecting a community board recommendation outcome, fields are getting validated when selecting the disapproved or Received after clock Expired option

![image](https://user-images.githubusercontent.com/57362229/69574375-f5680480-0f95-11ea-935d-c52847926b10.png)

trbmcginnis:
Looks like this is a recommendation code issue where there are certain codes (e.g. `717170002`) that mean different recommendation labels (e.g. 'Disapproved') for different participant types (e.g. 'Community Board' vs. 'Borough Board'). There's an action in `add.js` that does not specify that each of these different codes might mean something different for different participant types. Fixing now.

ksinghprod:
This issue is now fixed

PR #1109, opened 2020-02-20
1019: Fix field validation bug in recommendation form

trbmcginnis:
### Bug:
Previously, a user selection of the recommendation options `Disapproved` and `Received after Clock Expired` would cause the Votes in Favor, Votes Against, Abstain, and Total Members fields to automatically validate. This would happen without the user interacting with those fields or
clicking the "Continue" button, which is not expected behavior.

Addresses Issue #1019 

### Cause of Bug:
The bug was being caused by a recommendation code error in the action `setDispositionChangesetRec` in the `recommendations/add.js` controller. The action was checking for three recommendation codes (e.g. `717170002`) and assuming these all to correspond to the label `Waiver of Recommendation`. But different codes correlate to different labels based  on the participant type (e.g. `Community Board`).

See `RECOMMENDATION_OPTIONSET_BY_PARTICIPANT_TYPE_LOOKUP` in `recommendations/add.js` controller for these recommendation code-to-label relationships.

### Fix:
In order to fix this, the if statement checking for relevant recommendation codes in this action was broken out into two separate if statements, each one checking for codes that correlated to a specific participant type.

### DIFF
--- a/app/controllers/my-projects/assignment/recommendations/add.js
+++ b/app/controllers/my-projects/assignment/recommendations/add.js
@@ -270,8 +270,13 @@ export default class MyProjectsProjectRecommendationsAddController extends Contr
       dispositionChangeset.set(targetField, recommendation);
     }
 
-    // IDs for Waiver of Recommendation
-    if (([717170002, 717170006, 717170008].includes(recommendation)) && (participantType !== 'BP')) {
+    // For Borough Board, 71717002 = 'Waiver of Recommendation', For Community Board, 71717008 = 'Waiver of Recommendation'
+    const boroughBoardWaiverOfRecommendation = ([717170002].includes(recommendation)) && (participantType === 'BB');
+    const communityBoardWaiverOfRecommendation = ([717170008].includes(recommendation)) && (participantType === 'CB');
+
+    // A selection of 'Waiver of Recommendation' makes these four vote & member fields irrelevant
+    // Borough Presidents (BP) do not have these vote & member fields at all
+    if (boroughBoardWaiverOfRecommendation || communityBoardWaiverOfRecommendation) {
       dispositionChangeset.validate('dcpVotinginfavorrecommendation');
       dispositionChangeset.validate('dcpVotingagainstrecommendation');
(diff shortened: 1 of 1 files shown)


app/controllers/my-projects/assignment/recommendations/add.js
```diff
@@ -270,8 +270,17 @@ export default class MyProjectsProjectRecommendationsAddController extends Contr
       dispositionChangeset.set(targetField, recommendation);
     }
 
-    // IDs for Waiver of Recommendation
-    if (([717170002, 717170006, 717170008].includes(recommendation)) && (participantType !== 'BP')) {
+    // A selection of 'Waiver of Recommendation' makes these four vote & member fields irrelevant
+    // Borough Presidents (BP) do not have these vote & member fields at all
+    // For Borough Board, 71717002 = 'Waiver of Recommendation'
```
godfreyyeung: Since 277-280  seems to be a duplicate of 284-287, how about combining the conditionals? Something like:
```
if (
   ([717170008].includes(recommendation) && (participantType === 'CB')) ||
   ([717170002].includes(recommendation) && (participantType === 'BB'))
) {
   ...
}
```

allthesignals:
@trbmcginnis when you get back let's look at @godfreyyeung's comments :)

allthesignals:
@trbmcginnis we should wrap this up, any thoughts on @godfreyyeung comments?

app/controllers/my-projects/assignment/recommendations/add.js
trbmcginnis: Cool, I'll make the changes

allthesignals:
I'm so sorry but can you squash that last commit into the previous? If it's just a little lint fix it can easily go into there. Sorry we just wanna make sure our commits are clean :) thank youuuuu

trbmcginnis:
> I'm so sorry but can you squash that last commit into the previous? If it's just a little lint fix it can easily go into there. Sorry we just wanna make sure our commits are clean :) thank youuuuu

Whoops forgot to do that!

PR #1181, opened 2020-03-15
Rework for new backend API

allthesignals:
Don't panic


client/config/environment.js
```diff
@@ -143,7 +143,7 @@ module.exports = function(environment) {
 };
 
 function getHost(environment) {
-  if (process.env.HOST) {
+  if (process.env.HOST && environment === 'development') {
```
allthesignals, line 146: Hmmm this shouldn't work... and it's confusing. Rework this...

server/src/assignment/assignment.service.ts
```diff
@@ -98,7 +119,7 @@ export function transformIntoAssignments(projects, contactid) {
         // filter all dispositions so they're scoped to the user only
         // TODO: value mapping is making this lookup not work right, need to
         // find a better way to provide BOTH orig values and labeled vals
-        // .filter(disposition => disposition._dcp_recommendationsubmittedby_value === contactid)
+        .filter(disposition => disposition._isContactDisposition)
```
allthesignals, line 124: But also make sure all dispos are provided on the project too

Issue #1184, opened 2020-03-16
Filter counts are different when filtering by CD


hannahkates:
We should fix this when we do #1155

**Filter for:**
project community district = ____
AND project visibility = 'General Public'

allthesignals:
I'm not sure why this is. I'm using the dcp_communitydistricts field (I know we're updating that soon). it could be that it's not case insensitive? I'd be interested in seeing which ones are missing.

allthesignals:
This could also be that some records were missed again in the ETL transfer (yikes).

hannahkates:
Suggestion: look into "inactive" CDs

hannahkates:
This should be fixed by #1214 - will check on staging once the PR is merged.

PR #1214, opened 2020-06-06
1155: switch to use validated cds field

### LINKED ISSUE #1155: Update field source for Community Districts
AB#10109
[AB#10649](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/10649)

The Community District data attribute is used to 1) filter projects by CD and 2) to display CDs on a project profile.

Currently we use `dcp_communitydistricts` which is a string field containing 4 digit CD abbreviations delineated with semicolons. Ex: 'BK05;BK16;BK04'

The preferred field in CRM is changing to be `dcp_validatedcommunitydistricts`. The contents of this field are similar but comma separated and 3 digits. K00 (Brooklyn), R00 (SI), X00 (Bronx), M00 (Manhattan), Q00 (Queens). Ex: 'M01,X09,M04,R03,X06'

We need to update our app to:
- [x] Use the new correct field `dcp_validatedcommunitydistricts`
- [x] Use commas as the delimiter
- [x] Update any look ups that were dependent on the two digit borough abbreviations. Ex: BK -> K

hannahkates:
Addresses #1155 and #1184.

Updated to:
- [x] Use the new correct field `dcp_validatedcommunitydistricts`
- [x] Use commas as the delimiter
- [x] Update any look ups that were dependent on the two digit borough abbreviations. Ex: BK -> K
- [x] Remove references to other incorrect CD fields


client/app/models/project.js
```diff
@@ -75,10 +75,6 @@ export default class ProjectModel extends Model {
 
   @attr('string') dcpUlurpNonulurp;
 
-  @attr() dcpCommunitydistrict;
```
allthesignals, line 78: We're removing this because it's not used, right? J/c

client/mirage/factories/project.js
```diff
@@ -136,15 +136,7 @@ export default Factory.extend({
...
-  dcpCommunitydistrict() {
-    return faker.random.word();
-  },
-
-  dcpCommunitydistricts: 'BX08;SI01;SI02;SI03',
-
-  dcpValidatedcommunitydistricts() {
-    return faker.random.word();
-  },
+  dcpValidatedcommunitydistricts: 'X08;R01;R02;R03',
```
allthesignals, line 139: Should we update this to use the new format? It seems like validated are comma separated? Or no?

client/app/models/project.js
hannahkates: Correct, I there are three CD fields and CRM and I want to delete all references to the ones we're not using (dcp_communitydistrict and dcp_communitydistricts) to avoid confusion.

client/mirage/factories/project.js
hannahkates: Yes, good catch! Should be comma-separated. I'll fix.

hannahkates:
I tested this locally and it's working! Ready for review.

Issue #1158, opened 2020-03-05
Display "Prefiled" projects to LUPs on "Upcoming" tab


hannahkates:
[AB#10646](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/10646)

## User Story
LUPs need to be able to see "Prefiled" projects on the "Upcoming" tab so that they can prepare for this before the projects certify. However, "Prefiled" projects are tracked differently in the system and displaying them for LUPs will require a change to the LUP tab logic.

## Unresolved questions
What should the display for these "prefiled" projects look like on the upcoming tab? The current milestone timeline on the far right won't have any activity in it, so maybe some text should be displayed instead. Something along the lines of "This application has not yet been filed."

 ## Logic
In addition to the regular filed projects that are shown on the "Upcoming" tab, we need to include "Prefiled" projects which meet the criteria in the table below.

**In order to link the user to the project, we will need to use the dcp_validatedcommunitydistricts field to create a lookup that uses the contact's fullname**

![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/2d1514c1-738d-41e4-9428-eff658b7fb82)

Eabrante:
@hannahkates user story has been updated

andycochran:
I like the idea of replacing the milestone timeline with a sentence indicating that it's a "pre-filed" project: 

> This application has not yet been filed.

Or maybe: 

> Application expected to be filed within X days.

Let's check with the product owner to confirm what language to use.

bean75:
"Filing expected soon" if it needs to be short or "This application has not yet filed."

hannahkates:
Final logic

![Screen Shot 2020-06-30 at 1.24.42 PM.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/dd692fd4-1aa3-4c65-b705-ad76e7cb138c)

PR #1216, opened 2020-06-07
1158: include prefiled in lup upcoming

hannahkates:
This should be finished and reviewed AFTER #1214 is merged.

**Remaining to do:**
- Capture logged in contact's `fullname` for use in the assignments controller.
- Create a new WebAPI query for projects where 
  - `dcp_validatedcommunitydistricts` contains `recodedCbFullName`
  - `dcp_ulurp_nonulurp` = 'ULURP'
  - `dcp_publicstatus` = 'Prefiled'
- Union the results of that query ^ with the results of `const { records: projects } = await this.dynamicsWebApi.queryFromObject('dcp_projects', queryObject);`

**Front-end changes:**
- On the upcoming tab, if the project's `dcp_publicstatus` is "Prefiled", template hides the milestone timeline and says the application hasn't been filed yet.

**Back-end changes:**
- Creates a function that recodes the CB user's `fullname` to use the same CD abbreviations as `dcp_validatedcommunitydistricts` - making it possible to join a user with a project based on `dcp_validatedcommunitydistricts` instead of the LUP team assignment (Prefiled projects don't have LUP team assignments).
- Adds an "if" condition that checks if a project is ULURP, prefiled, and `dcp_validatedcommunitydistricts` contains the LUP user's name. This is only relevant for CBs.
- Captures `fullname` from the contact record so it can be used as an input to the assignment controller

Addresses #1158


server/src/assignment/assignment.controller.ts
```diff
@@ -33,13 +33,16 @@ export class AssignmentController {
...
       // todo: turn into CRM API
       if (email) {
-        ({ contactid } = await this.contactService.findByEmail(email));
+        const contact = await this.contactService.findByEmail(email);
+        const contactid = contact.contactid;
+        const fullname = contact.fullname;
       }
 
-      const records = await this.assignmentService.getAssignments(contactid, tab);
+      const records = await this.assignmentService.getAssignments(contactid, tab, fullname);
```
allthesignals, line 45: Use this for manual spot checking

allthesignals: line 36: need a way to find fullname

server/src/assignment/assignment.service.ts
```diff
@@ -228,6 +239,14 @@ function computeStatusTab(project, lupteam) {
     return 'archive';
   }
 
+  if (
+    project.dcp_publicstatus === 717170005 // Prefiled
+      && project.dcp_ulurp_nonulurp === 717170001 // ULURP
+      && project.dcp_validatedcommunitydistricts.includes(recodedCbFullName)) { // TODO
+    console.log('Prefiled!');
+    return 'upcoming';
+  }
```
allthesignals, line 248: The data for this is getting filtered out beforehand at the web api moment

allthesignals: Seems to have something to do with lupteammember doesn't have a team member assignment because they won't have that record

allthesignals: prefiled projects just aren't being included for some rason

Issue #1153, opened 2020-03-02
Revise LUP tab logic


hannahkates:
[AB#10645](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/10645)

Based on conversations with our Product Owner, we need to update the logic used to determine which projects show up on which tabs for LUPs. The preferred logic is captured in the image below. The tab should be determined based on the disposition visibility and status, removing all dependencies on milestone status.

![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/d3d5a6e2-3b03-4ccf-baae-f38fd8d774d9)

allthesignals:
This might be a good time to rewrite LUP portal read queries against the CRM API. I'll get a sense of how long that might take.

Eabrante:
@hannahkates user story has been updated

PR #1218, opened 2020-06-12
1153: revise logic for lup tab assignments

hannahkates:
Addresses #1153. The issue contains a table documenting the desired logic for placing projects on the different tabs.

This PR:
- Updates the logic for the tab assignments to depend on disposition visibility instead of milestone status
- Removes the visibility filter that was previously being applied to all dispositions

### DIFF
--- a/server/src/assignment/assignment.service.ts
+++ b/server/src/assignment/assignment.service.ts
@@ -2,7 +2,6 @@ import { Injectable } from '@nestjs/common';
 import { OdataService, overwriteCodesWithLabels } from '../odata/odata.service';
 import {
   all,
-  any,
   comparisonOperator,
   containsAnyOf
 } from '../odata/odata.module';
@@ -175,10 +174,6 @@ function generateAssignmentsQueryObject(query) {
     }),
   );
   const DISPOSITIONS_FILTER = all(
-    any(
-      comparisonOperator('dcp_visibility', 'eq', 717170004),
-      comparisonOperator('dcp_visibility', 'eq', 717170003),
-    ),
     `(not ${comparisonOperator('statuscode', 'eq', 717170001)})`,
   );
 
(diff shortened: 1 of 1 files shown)


server/src/assignment/assignment.service.ts
```diff
@@ -228,22 +215,35 @@ function computeStatusTab(project, lupteam) {
...
+      [717170003, 717170004].includes(disposition.dcp_visibility)
+      && ['Draft', 'Saved'].includes(disposition.statuscode)
+      && ['Active'].includes(disposition.statecode))
+    ) {
     return 'to-review';
   }
 
+  if (
+    participantDispositions.find(disposition =>
+      disposition.dcp_visibility == null // upcoming if dcp_visibility is null
```
godfreyyeung, line 241: nit: this comment can be removed, since the code is clear enough that it is self-documenting

server/src/assignment/assignment.service.ts
```diff
@@ -228,22 +215,35 @@ function computeStatusTab(project, lupteam) {
...
-  if (participantProjectMilestones.find(milestone => milestone.statuscode === 'Not Started')) {
-    return 'upcoming';
-  }
-
   // TODO: this is sensitive to sort order. we need to revise this!!!
   if (
-    participantProjectMilestones.find(milestone => ['In Progress', 'Completed'].includes(milestone.statuscode))
-      && participantDispositions.find(disposition => ['Inactive'].includes(disposition.statecode) && ['Submitted', 'Not Submitted'].includes(disposition.statuscode))) {
+    participantDispositions.find(disposition =>
+      // ['General Public', 'LUP'].includes(disposition.dcp_visibility)
```
godfreyyeung, line 221: We should remove these commented out lines. Or, if the intention was to clarify what 717170003 and 717170004 represent, we should change the comment to something like:
```
// 717170003 - General Public
// 717170004 - LUP
```
So that the comment isn't confused for commented out code

Issue #1219, opened 2020-07-06
Projects With Project visibility other than general public are displaying on the LUP portal 


ksinghprod:
## Browser
 Chrome 
## Project ID: 
2021M0102
Komal's Test Project for LUPP Enhancements
## Expected Result
The following logic should be met in order for Projects/dispositions to appear on the LUP Portal
Project Status = Active, On-Hold
Project Visibility = **General Public**  
Project Disposition State Code = Active
Signed in user is an Active LUP member on the project as CB, BP or BB  
Project milestone is =< Community Board Referral, Borough President Referral or Borough Board Referral
## Actual Result: 
Projects/dispositions records with applicant only, Internal DCP only, CPC only, LUP  visibility are displaying on the LUP Portal 

ADO User story: https://dcp-paperless.visualstudio.com/ZAP%20Portals/_workitems/edit/10645

![image](https://user-images.githubusercontent.com/57362229/86617759-c3843b00-bf85-11ea-92a0-6244d1bc1a6e.png)

![image](https://user-images.githubusercontent.com/57362229/86618725-6b4e3880-bf87-11ea-91f2-cfd027129853.png)

![image](https://user-images.githubusercontent.com/57362229/86617723-b2d3c500-bf85-11ea-897c-a0f259738eb8.png)

ksinghprod:
This issue is happening when Community districts are added to a project and then creeper mode is used to see that community board member;s projects. The project/disposition is appearing for project visibility other than general public

hannahkates:
Hi @ksinghprod just wanted to make sure you saw this is resolved.

ksinghprod:
This issue is now fixed 


![Media2](https://user-images.githubusercontent.com/57362229/88404081-6537c480-cd9b-11ea-843e-6a0fc904420b.gif)

PR #1220, opened 2020-07-06
1219: add visibility filter for lup assignments

hannahkates:
Addresses #1219.

Adds a filter to the Web API query that only gets projects where `dcp_visibility` is 'General Public' (71717003).

### DIFF
--- a/server/src/assignment/assignment.service.ts
+++ b/server/src/assignment/assignment.service.ts
@@ -214,14 +214,15 @@ function generateAssignmentsQueryObject(contact) {
     // filters by projects with associations having some connection to contactid
     // OR infers an association through recodedCbFullName (for prefiled)
     $filter: `
-      (dcp_dcp_project_dcp_communityboarddisposition_project/any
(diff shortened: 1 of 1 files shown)


server/src/assignment/assignment.service.ts
```diff
@@ -214,14 +214,15 @@ function generateAssignmentsQueryObject(contact) {
     // filters by projects with associations having some connection to contactid
     // OR infers an association through recodedCbFullName (for prefiled)
     $filter: `
-      (dcp_dcp_project_dcp_communityboarddisposition_project/any
+      dcp_visibility eq 717170003
```
godfreyyeung, line 217: We should add a comment as to what 717170003 means (what the human readable label is)

allthesignals: Ditto, we'll want to start adding these lookups so we can label stuff:

```javascript
  dcp_visibility eq ${ASSIGNMENT_VISIBILITY.PUBLIC}
```

Or something

Issue #1223, opened 2020-07-07
The "To Review Tab" is displaying hearings&recommendations without the LUP Disposition visibility are being general public/LUP


ksinghprod:
## Project Name
16th Avenue Rezoning 
2019K0024

## Expected Result:

![image](https://user-images.githubusercontent.com/57362229/86812646-3c01ff00-c04d-11ea-92a7-fe58e9318c2f.png)

## Actual Result: 

![image](https://user-images.githubusercontent.com/57362229/86812698-4ae8b180-c04d-11ea-9e29-77d8b1ecd71d.png)

![image](https://user-images.githubusercontent.com/57362229/86812853-79ff2300-c04d-11ea-927f-2ce4106d7848.png)

ksinghprod:
The same is happening for the "Reviewed Tab"

hannahkates:
Will be fixed by #1224

ksinghprod:
@hannahkates

 This issue is not fixed, my project that has Internal DCP only Visibility displayed in the "to be reviewed" tab. (with the disposition's visibility being general public)  

![image.png](https://images.zenhubusercontent.com/5ddc41dcbf921600010fc58b/2761c2ca-c6e9-4385-aeeb-7889210a8d23)

![image.png](https://images.zenhubusercontent.com/5ddc41dcbf921600010fc58b/6393d14d-bacd-4e57-a1e9-56c970a45fc9)

![image.png](https://images.zenhubusercontent.com/5ddc41dcbf921600010fc58b/38e4c823-1032-4d61-9e3a-f9729be30a21)

ksinghprod:
This issue is now fixed 

![image](https://user-images.githubusercontent.com/57362229/88410960-1a22af00-cda5-11ea-87b6-7d0292c8182b.png)


![image](https://user-images.githubusercontent.com/57362229/88410939-1131dd80-cda5-11ea-8e7c-ad1639a547c8.png)


![image](https://user-images.githubusercontent.com/57362229/88410986-29096180-cda5-11ea-890b-28a914c447d6.png)




![image](https://user-images.githubusercontent.com/57362229/88411020-3aeb0480-cda5-11ea-8e30-ffb97e9a3ec5.png)

Issue #1210, opened 2020-04-29
When there are multiple BP's/CB's/BB's on the LUP team, each member should see their own hearing/recommendation


hannahkates:
All LUPs assigned to a project should be able to see the project on their dashboard until they submit their recommendation, regardless of whether another LUP on that project has already submitted. 

In the current app implementation, it appears that the app is finding the first submitted disposition record on the project, and it assumes that ALL dispositions are already submitted for ALL assigned LUPs, when that isn't true.

**Steps to reproduce:**
- Assign yourself to a project that another LUP is already assigned to (one that would appear on the "to review" tab), making sure the disposition records are visible.
- Login to ZAP LUP as yourself, and see that the project shows up under "to review"
- Use creeper mode to see that the project also shows up under "to review" for the other LUP
- As yourself, submit your hearing and recommendation.
- See that the project moves off of your "to review" tab onto "reviewed"
- Use creeper mode to see that the project is now missing on "to review" for the other LUP, even though they haven't submitted their recommendation yet.
- In Dynamics, see that your disposition record is filled out and submitted, but the other LUP's isn't.

ksinghprod:
Hi @allthesignals @hannahkates you can try to use these repro steps,

 Assign 1 BP LUP to a project then submit the BP Recommendation (while keeping the BP review Milestone in Progress on CRM) then add another BP on the same project.
 
The Second BP will not see the option to submit the BP hearing/ recommendation, it will appear under the reviewed tab

ksinghprod:
This issue is now fixed

![image.png](https://images.zenhubusercontent.com/5ddc41dcbf921600010fc58b/902f07e1-fc3c-460f-87e4-d386e6485ae3)

![image.png](https://images.zenhubusercontent.com/5ddc41dcbf921600010fc58b/0265e7eb-3765-421a-b059-ac1ef77a7ad2)

PR #1224, opened 2020-07-09
1210: update computeStatusTab to base tab status off of only that use…

trbmcginnis:
Previously `computeStatusTab` was basing the tab status off of ALL dispositions that belonged to a project, making it so that when one user would submit a recommendation it would affect the tab status of all other user's assignments related to that project.

This PR filters the `participantDispositions` to only dispositions that are associated with that specific user. 

Addresses #1210

### DIFF
--- a/server/src/assignment/assignment.service.ts
+++ b/server/src/assignment/assignment.service.ts
@@ -21,7 +21,7 @@ export class AssignmentService {
     const { records: projects } = await this.dynamicsWebApi
       .queryFromObject('dcp_projects', queryObject);
 
-    return transformIntoAssignments(projects, contactid, recodedCbFullName)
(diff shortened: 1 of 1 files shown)


hannahkates:
Working great @trbmcginnis thank you! Added bonus, this fixed #1223 too.

server/src/assignment/assignment.service.ts
```diff
@@ -21,7 +21,7 @@ export class AssignmentService {
     const { records: projects } = await this.dynamicsWebApi
       .queryFromObject('dcp_projects', queryObject);
 
-    return transformIntoAssignments(projects, contactid, recodedCbFullName)
+    return transformIntoAssignments(projects, contactid, recodedCbFullName, fullname)
```
allthesignals, line 24: What is the difference between `recodedCbFullName` and `fullname`?

Issue #1221, opened 2020-07-06
LUP unable to submit hearing on "upcoming" when assigned as both BP and BB


hannahkates:
## What should happen
If an LUP user is assigned a BP and BB role on a project, they should be able to submit hearing info from the upcoming tab for each of the roles.

## What's happening instead
The project is showing up twice on the upcoming tab, reflecting the two different roles. However, the hearing button is only enabled for one of the assignment cards. The frontend `dispositionsByRole` attribute is what determines whether the button is enabled. It is null for one of the assignment cards.

Example project: 2021M0102
![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/ebfaa299-4035-4767-9594-9ffa33df88d6)

## Steps to reproduce
1. Find a test project to use. 2021M0102 is a good example.
2. Assign a user to the project LUP team as both a BB role and BP role. singh.komalpreet20%40gmail.com is already assigned this way for 2021M0102
3. View the user's upcoming tab on the LUP dashboard.
4. See that the project assignment is listed twice, but only one of the upcoming hearing buttons are enabled.
5. Use ember data inspector to see that `dispositionsByRole` is null for the assignment card with the disabled button.
6. Deactivate the LUP team member assignment that had the enabled button. Refresh LUP and see that the remaining assignment now contains data for `dispositionsByRole` and the button is enabled.

It seems like something may be wrong with the `dispositionsByRole` computed value in the assignments model. When there are multiple roles on a project for the same user, the app seems to only find disposition data for whichever assignment relationship was created most recently.

godfreyyeung:
This is an amazing issue writeup. Thank you @hannahkates !!!

hannahkates:
Hi @ksinghprod just wanted to make sure you saw this is resolved.

ksinghprod:
Hi Hannah, I thought I closed this, It is working. I re tested it the day you added the resolved label. Ill add a screenshot and close

ksinghprod:
This issue is now fixed 

![image](https://user-images.githubusercontent.com/57362229/88410326-3eca5700-cda4-11ea-904d-c0b8775157c0.png)

PR #1225, opened 2020-07-09
1221: filter dispositions based on dcp_representing in backend

trbmcginnis:
If ONE user had two roles (BB and BP) and therefore two assignments, one for each role, ALL dispositions (regardless of role) were nested on BOTH assignments. This was because we were only filtering assignment.dispositions based on contact id rather than on both contact id and role. 

This PR updates this filtering logic to include the role associated with the assignment as well. 

Addresses #1221

### DIFF
--- a/server/src/assignment/assignment.service.ts
+++ b/server/src/assignment/assignment.service.ts
@@ -114,11 +114,16 @@ export function transformIntoAssignments(projects, contactid, recodedCbFullName)
       const actions = transformActions(project.dcp_dcp_project_dcp_projectaction_project);
       const milestones = project.dcp_dcp_project_dcp_projectmilestone_project;
       const dispositions = project.dcp_dcp_project_dcp_communityboarddisposition_project;
+      const participantRoleLookup = {
+        BP: 'Borough President',
+        BB: 'Borough Board',
+        CB: 'Community Board',
+      };
       const userDispositions = dispositions
         // filter all dispositions so they're scoped to the user only
         // TODO: value mapping is making this lookup not work right, need to
         // find a better way to provide BOTH orig values and labeled vals
-        .filter(disposition => disposition._isContactDisposition)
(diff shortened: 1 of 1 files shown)


server/src/assignment/assignment.service.ts
```diff
@@ -114,11 +114,16 @@ export function transformIntoAssignments(projects, contactid, recodedCbFullName)
       const actions = transformActions(project.dcp_dcp_project_dcp_projectaction_project);
       const milestones = project.dcp_dcp_project_dcp_projectmilestone_project;
       const dispositions = project.dcp_dcp_project_dcp_communityboarddisposition_project;
+      const participantRoleLookup = {
+        BP: 'Borough President',
+        BB: 'Borough Board',
+        CB: 'Community Board',
+      };
```
allthesignals, line 121: Nit but I'd always keep lookups like this in the very top of the file and with ALL_CAPS:

```javascript
      const PARTICIPANT_ROLE_LOOKUP = {
        BP: 'Borough President',
        BB: 'Borough Board',
        CB: 'Community Board',
      };
```

server/src/assignment/assignment.service.ts
```diff
@@ -114,11 +114,16 @@ export function transformIntoAssignments(projects, contactid, recodedCbFullName)
...
       const userDispositions = dispositions
         // filter all dispositions so they're scoped to the user only
         // TODO: value mapping is making this lookup not work right, need to
         // find a better way to provide BOTH orig values and labeled vals
-        .filter(disposition => disposition._isContactDisposition)
+        .filter(disposition => disposition._isContactDisposition && disposition.dcp_representing === participantRoleLookup[lupteam.dcp_lupteammemberrole])
```
allthesignals, line 126: Great job figuring this out. Where does `_isContactDisposition` get set? It might have been clearer to make another property like that for this condition.

PR #1234, opened 2020-08-24
12122 Fix defaultValue for nameofpersoncompletingthisform

godfreyyeung:
The Disposition model's dcpNameofpersoncompletingthisform
property was not being included in request payloads
to the backend. The fix is to remove the
`string-null-triggers-default` Ember Data transform,
and instead set up a pipe on the dispostion PATCH
endpoint to provide defaults/transform values.

Todo: standardize usage of pipes for setting default
values for incoming attributes.

Fixes [AB#12122](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12122)


server/src/disposition/defaultValue.disposition.pipe.ts
```diff
@@ -0,0 +1,14 @@
+import { PipeTransform, Injectable, ArgumentMetadata } from '@nestjs/common';
+
+@Injectable()
+export class defaultValueDispositionPipe implements PipeTransform<any, any> {
```
allthesignals, line 4: The thing about pipes is that they're sorta generic, sorta not. I don't have a problem with this but because these pipes can be decorated onto anything, regardless of whether it's a disposition class, makes the naming here confusing.

Still, I like this approach, but I would keep that in mind.

godfreyyeung: Yeah, good point. I was also thinking of extracting out the default values into a "default values" file, and parameterizing the pipe. What do you think?

godfreyyeung: the pipe would be placed into a new folder like `/pipe`. Like
`server/src/pipe/defaultValue` 

And then the default value file would be specific to disposition:
`server/src/disposition/default-value.ts`

allthesignals: That would be cool. I didn't know you could apply decorators in taht way.

My concern is that it's method-specific

allthesignals: For example: why is the defaultValue not applied on the get requests?

godfreyyeung: Good Q. We definitely could... In this case the property "dcpNameofpersoncompletingthisfield" is never displayed or edited by the user, so it doesn't matter as much. But otherwise I think there's a bit of design thought involved, and just agreeing as a team to transform certain backend values before they reach the frontend

PR #1245, opened 2020-11-05
13242 Remove fields and search options from FE/API

CeciliaCY:
### Summary
13242 Remove fields and search options from FE/API

Close [AB#13242](https://dcp-paperless.visualstudio.com/ZAP%20Portals/_workitems/edit/13242) 

### Technical Explanation
Remove all items related to dcpFemafloodzonecoastala and   dcpFemafloodzonev


CeciliaCY:
@allthesignals could you please check my code, the front end is working as expected, but I don't know if there is any other tests we can do. Thanks!

godfreyyeung (approved):
That looks good to me, but probably better if @allthesignals  gives a look too!

PR #1246, opened 2020-11-06
13264 Restore map on project profile page

godfreyyeung:
### Summary
This PR restores the map on the Project profile page. Previously it was disabled and read "No map available"
image

Fixes [AB#13264](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/13264)

### Technical Details
The changes are mostly backend because the crux of the fix is to ensure that a value for the Project bblFeaturecollection property is sent from the backend.

The backend took a couple steps to construct the bblFeaturecollection value
  1. When constructing the response for a single project (project.service.ts -> `findOneByName()`), sends a Fetch request to the Carto API to get Mappluto geometry data for the target Project's list of BBLs. 
     - The Fetch request included a SQL query that made use of PostGresSQL's `st_AsGeoJSON` spatial function to transform geometry to GeoJSON. 
  2. Transforms the returned geometry GeoJSON for a polygon into a GeoJSON  "FeatureCollection", which is the expected format for Project.bblFeaturecollection in the frontend. This is specified in the sample test data file `client/mirage/test-data/bbl-feature-collection.js`

The frontend was already all set up to display the geometries again, once bblFeaturecollection is saturated!

### DIFF
--- a/server/src/carto/carto.service.ts
+++ b/server/src/carto/carto.service.ts
@@ -1,0 +1,44 @@
+import { Injectable } from '@nestjs/common';
+import * as fetch from 'node-fetch';
+
+@Injectable()
+export class CartoService {
+  cartoUsername = 'planninglabs';
+
+  buildSqlUrl(cleanedQuery, format = 'json', method) { // eslint-disable-line
+    let url = `https://${this.cartoUsername}.carto.com/api/v2/sql`;
+    url += method === 'get' ? `?q=${cleanedQuery}&format=${format}` : '';
+
+    return url;
+  };
+
+  async fetchCarto(query, format = 'json', method = 'get') {
+    const cleanedQuery = query.replace('\n', '');
+    const url = this.buildSqlUrl(cleanedQuery, format, method);
+
+    let fetchOptions = {};
+
+    if (method === 'post') {
+      fetchOptions = {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/x-www-form-urlencoded; charset=UTF-8',
+        },
+        body: `q=${cleanedQuery}&format=${format}`,
+      };
+    }
+
+    let cartoResponse = await fetch(url, fetchOptions);
+
+    if (cartoResponse.ok) {
+      cartoResponse = await cartoResponse.json();
+
+      if (format === 'json')
+        return cartoResponse.rows;
+
+      return cartoResponse;
+    }
+
+    throw new Error('Request to carto failed.');
+  };
+}
--- a/server/src/project/project.service.ts
+++ b/server/src/project/project.service.ts
@@ -2,6 +2,7 @@ import { Injectable } from '@nestjs/common';
 import { Serializer } from 'jsonapi-serializer';
 import { dasherize } from 'inflected';
 import { ConfigService } from '../config/config.service';
+import { CartoService } from '../carto/carto.service';
 import { handleDownload } from './_utils/handle-download';
 import { transformMilestones } from './_utils/transform-milestones';
 import { transformActions } from './_utils/transform-actions';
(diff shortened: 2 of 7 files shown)


server/src/project/project.service.ts
```diff
@@ -283,8 +284,50 @@ export class ProjectService {
   constructor(
     private readonly dynamicsWebApi: OdataService,
     private readonly config: ConfigService,
+    private readonly carto: CartoService,
   ) {}
 
+  async getBblsGeoJSON(bbls) {
```
godfreyyeung, line 290: Maybe I should rename as "getBblsGeometry"?

server/src/project/project.service.ts
```diff
@@ -283,8 +284,50 @@ export class ProjectService {
...
+    let collectedBBLs = bbls;
```
allthesignals, line 291: I know this is code from old commits, but I would avoid re-naming the argument like this haha. Or just rename the arg. Idk.

server/src/project/project.service.ts
allthesignals: That works too

server/src/project/project.service.ts
```diff
@@ -283,8 +284,50 @@ export class ProjectService {
...
+
+    if (collectedBBLs === null) return null;
+
+    collectedBBLs = collectedBBLs.filter(bbl => bbl !== null);
```
allthesignals, line 295: I'd Make the arg in teh func signature `bbls = []` with a default.

allthesignals: why are there null BBLs at all?

server/src/project/project.service.ts
```diff
@@ -283,8 +284,50 @@ export class ProjectService {
...
+    const cartoResponse = await this.carto.fetchCarto(SQL, 'json', 'post');
+
+    if (cartoResponse.length === 0) {
+      return {
+        polygons: null,
+      };
+    }
+
+    // return first object in carto response, carto.sql always return an array
+    return cartoResponse[0];
```
allthesignals, line 314: Is there a way we can move some of this munging stuff into the carto service? It seems like the responsibility of that service. could also request `geojson` — I think that return a less awkward structure?

server/src/project/project.service.ts
```diff
@@ -283,8 +284,50 @@ export class ProjectService {
...
+
+  async getBblsFeaturecollection(bbls) {
+    const { polygons: bblsGeoJSON } = await this.getBblsGeoJSON(bbls);
+
+    return {
+      type: 'FeatureCollection',
+      features: [
+        {
+          type: 'Feature',
+          geometry: JSON.parse(bblsGeoJSON),
```
allthesignals, line 325: Wait, this has to be parsed to JSON? This should definitely happen in the carto service, no?

server/src/carto/carto.service.ts
```diff
@@ -0,0 +1,42 @@
...
+    return fetch(url, fetchOptions)
+      .then((response) => {
+        if (response.ok) {
+          return response.json();
+        }
+        throw new Error('Not found');
+      })
+      .then((d) => { // eslint-disable-line
+        return format === 'json' ? d.rows : d;
+      });
```
allthesignals, line 40: Can we change this to async/await rather than classic promise methods?

allthesignals (approved):
Looks pretty good, had some optinoal thoughts on organizing things a little differently. Ur call. Thank you so much!! So happy to have map back

server/src/carto/carto.service.ts
godfreyyeung: Ahh good catch!! Will do

server/src/project/project.service.ts
godfreyyeung: Hmm, yeah maybe the Carto service can provide a utilitly function for this transformation. I supposed that transforming into a FeatureCollection is not necessarily carto-related, per se. But it would be nice to encapsulate that so consumers have less legwork and details to worry about..

server/src/project/project.service.ts
godfreyyeung: Good idea -- will look into it

server/src/project/project.service.ts
godfreyyeung: Good idea on the default args. Hmm yeah good q... That was a copy from old code, maybe the original author noticed sometimes null was included in the bbls array? not sure. I haven't seen it myself. We could remove it for now and watch out for edge cases

server/src/project/project.service.ts
godfreyyeung: I had that thought too. But collectedBBLs is modified on the fly a few lines down. It's not so much a renaming as it is assigning a default value to collectedBBLs

godfreyyeung: Though actually on second read it looks unnecessary. collectedBBs can just be declared on line 295 instead. 
Will change

server/src/project/project.service.ts
godfreyyeung: Hmm actually moving `[0]` into the service would break the abstraction/utility of the carto service, since I think it's possible there's more than one result for certain SQL queries.

PR #1247, opened 2020-11-09
13238 Set up backend to deliver Project Package docs

godfreyyeung:
### Summary
This PR sets up the backend so that it sideloads an individual Project with Packages, and each package has an array of documents. This is in preparation to list documents by package on the Profile Profile page. 

### Task
Fixes [AB#13238](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/13238)

### Technical Details
  - The entry point for this PR is the `getProject()` method in project.service.ts. There, it will acquire the Project's associated packages, saturate it with documents, and insert it into the response payload for an individual project. 
  - The Applicant Portal's CRM service was (reverse?) ported to Zap Search, in hopes that it will "Strangle out" Zap Search's older API to talk to CRM. crm.service.ts is essentially a copy of what is currently in applicant portal.
  - The new document.service.ts is an improvement over the Applicant Portal's document.service.ts. Eventually some of the modifications could be moved over to Applicant Portal. 
  - I'm adding the `@Get('/*')` endpoint here in preparation for making the files available for download. @allthesignals  do you foresee any issue with this?

### DIFF
--- a/server/src/project/project.service.ts
+++ b/server/src/project/project.service.ts
@@ -11,6 +11,7 @@ import { injectSupportDocumentURLs } from './_utils/inject-supporting-document-u
 import { getVideoLinks } from './_utils/get-video-links';
 import { KEYS as PROJECT_KEYS, ACTION_KEYS, MILESTONE_KEYS } from './project.entity';
 import { KEYS as DISPOSITION_KEYS } from '../disposition/disposition.entity';
+import { PACKAGE_ATTRS } from '../package/packages.attrs';
 import { Octokit } from '@octokit/rest';
 import { OdataService, overwriteCodesWithLabels } from '../odata/odata.service';
 import {
@@ -24,6 +25,11 @@ import {
   equalsAnyOf,
   containsAnyOf
 } from '../odata/odata.module';
+// CrmService is a copy of the newer API we built for Applicant Portal to talk to CRM.
+// It will eventually strangle out OdataService, which is an older API to accomplish
+// the same thing.
+import { CrmService } from '../crm/crm.service';
+import { DocumentService } from '../document/document.service';
 
 const ITEMS_PER_PAGE = 30;
 const BOROUGH_LOOKUP = {
@@ -102,6 +108,13 @@ const FIELD_LABEL_REPLACEMENT_WHITELIST = [
   '_dcp_action_value',
   '_dcp_zoningresolution_value',
 ];
+
+const PACKAGE_VISIBILITY = {
+  GENERAL_PUBLIC: 717170003,
+}
+const PACKAGE_STATUSCODE = {
+  SUBMITTED: 717170012,
+}
 
 // configure received params, provide procedures for generating queries.
 // these funcs do not get called unless they are in the query params.
@@ -285,6 +298,8 @@ export class ProjectService {
     private readonly dynamicsWebApi: OdataService,
     private readonly config: ConfigService,
     private readonly carto: CartoService,
+    private readonly crmService: CrmService,
+    private readonly documentService: DocumentService,
   ) {}
 
   async getBblsGeometry(bbls = []) {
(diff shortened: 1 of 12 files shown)


server/src/project/project.service.ts
```diff
@@ -395,9 +411,26 @@ export class ProjectService {
...
+            dcp_visibility eq ${PACKAGE_VISIBILITY.GENERAL_PUBLIC}
+          )
+          and (
+            statuscode eq ${PACKAGE_STATUSCODE.SUBMITTED}
+          )
+      `);
+
+    projectPackages = await Promise.all(projectPackages.map(async (pkg) => {
+        return await this.documentService.packageWithDocuments(pkg);
+      }));
```
allthesignals, line 427: How does this perform? I think it's probably okay, but I could see these getting really slow. Have you considered getting the packages in the very first request? We're removing the expansion for keyword entities anyway, so there'll be room for another one.

I wonder if there's a way to get a filtered list of folders in SharePoint based on project name, and also get a list of files there, in a single request? I wouldn't spin your wheels on it too much, but I have a hard time seeing this perform very well. Another thing we could do to prevent a bottle neck is to make the entire files/artifacts section asynchronous from the front end: so, have another component make a separate request to the API to a dedicated endpoint that does this separately.

server/src/project/project.service.ts
```diff
@@ -285,6 +298,8 @@ export class ProjectService {
     private readonly dynamicsWebApi: OdataService,
     private readonly config: ConfigService,
     private readonly carto: CartoService,
+    private readonly crmService: CrmService,
```
allthesignals, line 301: I mean, I think it's okay to just crib the service from applicant portal, but with OdataService still there, it seems to make more sense to just copy the public methods you need. Just be aware of the lack of tests.

allthesignals (commented):
> Do you think there's any security concern with this? Like could people use it to download files that are not supposed to be public, if they guess the right url? 

Yes, people would be able to download the file if they guessed the right URL. But that's only for the folder/filename, and it would affect packages and revisions that are not public.

There could be a check for that, making sure the package (which could be inferred from the folder name) is public. Could also do some fancy stateless stuff with jwt's.

Let's see how well this works and build from there.

server/src/project/project.service.ts
```diff
@@ -395,9 +411,26 @@ export class ProjectService {
...
+    projectPackages = await Promise.all(projectPackages.map(async (pkg) => {
+        return await this.documentService.packageWithDocuments(pkg);
+      }));
+
+    transformedProject.packages = projectPackages;
+
     await injectSupportDocumentURLs(transformedProject);
 
-    return this.serialize(transformedProject);
+    return this.serialize(transformedProject);;
```
godfreyyeung, line 433: remove extra semicolon here

server/src/project/project.service.ts
godfreyyeung: Hmm. Good idea, conceptually speaking including it in the initial expand might be faster.

It's been pretty speedy so far. I think that's because we are only loading document metadata (name, url, etc. instead of the documents themselves). I've only been testing on projects with 1-2 packages and 1-2 documents per package, but I think in terms of absolute response time, even with a dozen packages and dozen documents shouldn't decrease performance too dramatically since it's mostly text results. 

Interesting idea about getting a flat list of Project files from Sharepoint. After this PR I have to rework this code to read the sharepoint url from the package, anyway. I can look into that sort of thing..

server/src/project/project.service.ts
godfreyyeung: Right. There are some "delete" methods in there that aren't used so far... but I figured why not crib it over in preparation for future work. hmm 🤔

server/src/project/project.service.ts
allthesignals: >  I think that's because we are only loading document metadata (name, url, etc. instead of the documents themselves).

Yep, I know. But it's still a new network request for every package.

Just keep in mind there is a rate limit in Microsoft's API so we'd wanna be mindful. I could see how much we're eating up with the public-facing site.

godfreyyeung:
Prefer latest commit in #1248

PR #1248, opened 2020-11-10
13238 project packages documents

godfreyyeung:
### Summary
Sets up backend to deliver Projects with submitted, public Packages that have a sideloaded Documents information. 
Sets up backend code to deliver the Documents for download.
Sets up frontend code to list those Documents grouped by Packages on the Project profile page.

See project 2021M0268 in (on dev/uat2) for an example. 
image

#### Task/Bug Number
Fixes [AB#13238](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/13238)

### Technical Explanation
  - Introduced a guard to make sure only documents belonging to General Public, Submitted packages can be downloaded.
  - There's also the addition of a few extra guards to protect against errors when variables are null. Also some fixes due to problems revealed by deepscan (e.g. using regex for find and replace instead of a string).

### DIFF
--- a/server/src/document/document.controller.ts
+++ b/server/src/document/document.controller.ts
@@ -1,4 +1,5 @@
 import { Controller,
+  Get,
   Post,
   Req,
   Res,
@@ -6,9 +7,11 @@ import { Controller,
   UploadedFile,
   HttpStatus,
   HttpException,
+  Param,
   Session,
 } from '@nestjs/common';
 import { ConfigService } from '../config/config.service';
+import { DocumentService } from './document.service';
 import { FileInterceptor } from '@nestjs/platform-express';
 import { Request } from 'express';
 import { CRMWebAPI } from '../_utils/crm-web-api';
@@ -17,6 +20,7 @@ import { CRMWebAPI } from '../_utils/crm-web-api';
 export class DocumentController {
   constructor(
     private readonly config: ConfigService,
+    private readonly documentService: DocumentService,
   ) {}
 
   /** Uploads a single document
@@ -64,5 +68,12 @@ export class DocumentController {
     } else {
       response.status(400).send({ "error": 'You can only upload files to dcp_communityboarddisposition at this time' });
     }
+  }
+
+  @Get('/*')
+  async read(@Param() path, @Res() res) {
+    const stream = await this.documentService.getPublicPackageDocument(path);
+
+    stream.pipe(res);
   }
 }
--- a/server/src/document/document.service.ts
+++ b/server/src/document/document.service.ts
@@ -1,0 +1,173 @@
+import { Injectable, HttpStatus, HttpException } from '@nestjs/common';
+import { CrmService } from '../crm/crm.service';
+
+const PACKAGE_VISIBILITY = {
+  GENERAL_PUBLIC: 717170003,
+}
+const PACKAGE_STATUSCODE = {
+  SUBMITTED: 717170012,
+}
+
+// Inserts hyphens into an unhyphenated GUID string
+// e.g. 
+// 996699F37323EB11A813001DD8309FA8 
+// becomes
+// 996699F3-7323-EB11-A813-001DD8309FA8
+//
+// https://docs.microsoft.com/en-us/dynamics-nav/guid-data-type
+function hyphenateGUID(unhyphenatedGUID) {
+  return [
+    unhyphenatedGUID.slice(0, 8),
+    '-',
+    unhyphenatedGUID.slice(8, 12),
+    '-',
+    unhyphenatedGUID.slice(12,16),
+    '-',
+    unhyphenatedGUID.slice(16,20),
+    '-',
+    unhyphenatedGUID.slice(20),
+  ].join('');
+}
+
+@Injectable()
+export class DocumentService {
+  constructor(
+    private readonly crmService: CrmService,
+  ){}
+
+  // We retrieve documents from the Sharepoint folder (`folderIdentifier` in the
+  // code below) that holds both documents from past revisions and the current
+  // revision. CRM automatically carries over documents from past revisions into
+  // this folder, and we deliberately upload documents for the latest/current
+  // revision into this folder.
+  async findPackageSharepointDocuments(packageName, id: string) {
+    try {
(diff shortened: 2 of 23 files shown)


server/src/document/document.controller.ts
```diff
@@ -65,4 +76,80 @@ export class DocumentController {
...
+      strippedPackageId.slice(0, 8),
+      '-',
+      strippedPackageId.slice(8, 12),
+      '-',
+      strippedPackageId.slice(12,16),
+      '-',
+      strippedPackageId.slice(16,20),
+      '-',
+      strippedPackageId.slice(20),
+    ].join('');
```
allthesignals, line 119: This kind of stuff should at least go in a different function if not the service?

godfreyyeung: true!! can do

allthesignals: Make a function outside the class called "formatPackageToHyphenated" or something. Use your comments to make a function with that sort of name.

server/src/document/document.controller.ts
```diff
@@ -65,4 +76,80 @@ export class DocumentController {
...
     }
   }
+
+  // "path" refers to the "relative server path", the path
+  // to the file itself on the sharepoint host. It has the format
+  //
+  //   /sites/<site>/<entity_name>/<folder_name>/filename.pdf
+  //
+  // For example,
+  //
```
godfreyyeung, line 86: nit: delete extra blank line here

server/src/document/document.controller.ts
allthesignals: What do we know about these kinds of GUIDs? Do they follow a standard? Maybe there's the equivalent of something like "dasherize"

godfreyyeung: True they might follow a standard. It's been pretty consistent that they have the hyphens in those specific indexes. 
According to Microsoft docs:
https://docs.microsoft.com/en-us/dynamics-nav/guid-data-type

```
The GUID is a 16-byte binary data type that can be logically grouped into the following subgroups:

4byte-2byte-2byte-2byte-6byte.

The standard textual representation is {12345678-1234-1234-1234-1234567890AB}.
```

So yeah maybe there is a util already.. i can check

server/src/document/document.controller.ts
```diff
@@ -65,4 +76,80 @@ export class DocumentController {
...
+      const stream = await this.crmService.getSharepointFile(pathSegment);
+
+      stream.pipe(res);
+    } catch(e) {
+      const errorMessage = `Unable to provide document access.`;
+      throw new HttpException({
+        "code": "DOCUMENT_GET_ERROR",
+        "detail": errorMessage,
+      }, HttpStatus.FORBIDDEN);
+    }
```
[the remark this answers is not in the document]
allthesignals: i would look critically at how errors have been working for us in applicant portal and how we should improve on that. I have personally found it difficult to debug errors because the actual error message has been missing or because the error message falsely describes what actually happeened.

server/src/project/project.service.ts
```diff
@@ -395,6 +411,23 @@ export class ProjectService {
...
+          and (
+            dcp_visibility eq ${PACKAGE_VISIBILITY.GENERAL_PUBLIC}
+          )
+          and (
+            statuscode eq ${PACKAGE_STATUSCODE.SUBMITTED}
+          )
+      `);
+
+    projectPackages = await Promise.all(projectPackages.map(async (pkg) => {
+        return await this.documentService.packageWithDocuments(pkg);
```
allthesignals, line 426: is there a way to double check that certian packages should be making this query? perhaps we could check that the pkg is public? that would spare us some network requests.

allthesignals (approved):
approving but lots of organizational concerns.

server/src/document/document.controller.ts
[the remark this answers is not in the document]
godfreyyeung: My original thought is to, on principle, hide those CRM error details to the client/user and only log it in the backend (I should have included console.logs here). Though I think it in our case since it's an alpha/beta app and there's such high value for us to directly see the CRM errors in the client, that would make sense to include `e.message` in the exception payload.  

I'll include it here to get us started.
I'll create a task for us to work out Zap Search frontend/backend error reporting

server/src/document/document.controller.ts
```diff
@@ -65,4 +76,80 @@ export class DocumentController {
...
+  // of hyphens.
+  //
+  // We have assurance of this. See
+  // https://dcp-paperless.visualstudio.com/dcp-paperless-dynamics/_workitems/edit/13366
+  @Get('/*')
+  async read(@Param() path, @Res() res) {
+    const pathSegment = path[0];
+    const packageFolder = pathSegment.split('/')[3];
+    const packageFolderSegments = packageFolder.split('_');
+    const strippedPackageId = packageFolderSegments[packageFolderSegments.length - 1];
```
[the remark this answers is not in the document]
godfreyyeung: I wrote all this from scratch just now, but wouldn't be surprised if it mirrored some old code since the process is the same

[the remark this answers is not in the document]
godfreyyeung: yep i'll pull it out in to a helper

server/src/document/document.service.ts
```diff
@@ -0,0 +1,80 @@
+import { Injectable, HttpStatus, HttpException } from '@nestjs/common';
+import { CrmService } from '../crm/crm.service';
+
+@Injectable()
+export class DocumentService {
```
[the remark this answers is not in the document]
godfreyyeung: this is lifted, yep!

server/src/project/project.service.ts
godfreyyeung: yeah, the code above (414 - 423) limits the sideloaded packages to only those that are public and submitted. Other packages are not included here.

server/src/document/document.controller.ts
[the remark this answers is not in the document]
allthesignals: yeah when i say janky i just mean it's a weird pattern for the folder name. i think i had to do it in one of those methods for get docs

PR #1249, opened 2020-11-21
13237 List public artifact documents

godfreyyeung:
### Summary
Lists public documents from Artifacts on a project profile.

### Task
Fixes [AB#13237](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/13237)

### Technical Details
  - Follows same pattern previously used for docs from packages
  - Abstracts some of the path formatting into utility functions
  - Deletes unused tests

### Screens
image
image


allthesignals:
@godfreyyeung lemme know if I can help with this!

server/src/document/document.service.ts
```diff
@@ -74,6 +89,40 @@ export class DocumentService {
...
+      } else {
+        throw new HttpException({
+          code: 'SHAREPOINT_FOLDER_ERROR',
+          title: 'Bad Sharepoint folder lookup',
+          detail: `An error occured while constructing and looking up folder for artifact. Perhaps the artifact name or id is wrong.`,
+          meta: {
+            artifactName: artifactName,
+            artifactsId: id,
+          }
+        }, HttpStatus.INTERNAL_SERVER_ERROR);
```
allthesignals, line 121: Again, where is the inclusion of `e.message`? You use a template literal on 116, but no string interpolation.

Again, the biggest issue I've had with these errors, including in applicant portal, are how they conceal the actual error message. "An error occured" — okay, but what's the error? It's right there in the catch block argument.

I would think more critically about your own experience using this error handling system. Has it been frustrating? What could make it better?

server/src/document/document.service.ts
```diff
@@ -106,6 +155,38 @@ export class DocumentService {
...
+
+    try {
+      return {
+        ...projectArtifact,
+        documents: await this.findArtifactSharepointDocuments(dcp_name, dcp_artifactsid),
+      };
+    } catch(e) {
+      const errorMessage = `Error loading documents for artifact ${dcp_name}.`;
+      console.log(errorMessage);
+      console.log(e)
```
allthesignals, line 180: See comment above on error messages.

server/src/document/document.service.ts
```diff
@@ -151,13 +232,29 @@ export class DocumentService {
...
+      // So if firstPackage is undefined, we then attempt to retrieve an
+      // Artifact
       if (!firstPackage) {
-        const errorMessage = `No document access.`;
-        console.log(`Client attempted to retrieve document ${pathSegment}, but no associated public, submitted packages were found.`);
-        throw new HttpException({
-          "code": "NO_DOCUMENT_ACCESS",
-          "detail": errorMessage,
-        }, HttpStatus.FORBIDDEN);
+        const { records: [firstArtifact] } = await this.crmService.get('dcp_artifactses', `
```
allthesignals, line 240: Oh my good god, "dcp_artifactses"?! The naming just get better and better!

server/src/document/document.service.ts
```diff
@@ -151,13 +232,29 @@ export class DocumentService {
...
+
+        // If the record GUID also returns no Artifact, then there may be
+        // something wrong with the GUID 
+        if (!firstArtifact) {
+          const errorMessage = `No document access.`;
+          console.log(`Client attempted to retrieve document ${pathSegment}, but no associated public, submitted packages or artifacts were found.`);
+          throw new HttpException({
+            "code": "NO_DOCUMENT_ACCESS",
+            "detail": errorMessage,
+          }, HttpStatus.FORBIDDEN);
```
allthesignals, line 256: Not sure that this should imply a 403. It seems like a 404? In what situation would it be requesting something that isn't found?

allthesignals (approved):
@godfreyyeung lemme know if I can help with this PR

client/app/components/show-project/public-documents.js
```diff
@@ -6,12 +6,17 @@ export default class PublicDocumentsComponent extends Component {
   @tracked showPublicDocuments = false;
 
   get existsPublicDocuments () {
```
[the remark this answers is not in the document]
godfreyyeung: Good catch. I think I blanked when trying to think of the naming convention for this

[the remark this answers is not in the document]
godfreyyeung: I changed this for the similarly-named getters for artifact and package

client/app/components/show-project/public-documents/package-documents.js
```diff
@@ -82,6 +82,12 @@ export default class PackageDocumentsComponent extends Component {
     return 'Unknown package';
   }
 
+  get existsPackageDocuments() {
+    
+    console.log(this.package.documents);
```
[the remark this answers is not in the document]
godfreyyeung: Thx!

server/src/document/document.service.ts
godfreyyeung: Ah, good catch.. should have included that. I updated the exception detail to include `e`.
I decided to log and include the entire, `JSON.stringify()"-ed `e` because it sometimes contains a `code` property which might be valuable to debugging. 

I also updated the other functions to make it consistent across the service 

Yeah. I think it needs a concentrated effort to improve it,  so am going to dedicate it to a separate task.

server/src/document/document.service.ts
godfreyyeung: Yuuuuuup. This has also happened before with some of the form entities, like `subjectsites` I think. 🙈

server/src/document/document.service.ts
allthesignals: Very interesting, thanks for looking that stuff up @godfreyyeung

PR #1254, opened 2020-12-07
13239 display review meeting milestone use correct date field

dhochbaum-dcp:
#AB13239


server/src/project/_utils/transform-milestones.ts
```diff
@@ -146,6 +146,21 @@ function applyDisplayDate(milestone, project) {
   if (milestone._dcp_milestone_value === '780593bb-ecc2-e811-8156-1458d04d0698') milestone.display_date = milestone.dcp_actualenddate;
   if (milestone._dcp_milestone_value === '483beec4-dad0-e711-8116-1458d04e2fb8') milestone.display_date = milestone.dcp_actualenddate;
   if (milestone._dcp_milestone_value === '4a3beec4-dad0-e711-8116-1458d04e2fb8') milestone.display_date = milestone.dcp_actualenddate;
+  
+  /*
+  * The following is for task 13239
+  * The date displayed should be the CRM Review Meeting date for certain milestones if the milestone is "in progress" or "completed", and the meeting date is <= current date
+  * I have included a check to make sure that the field is not null to make sure we are not overwriting dates with nothing
+  * 2 is completed, 717170000 is in progress
```
godfreyyeung, line 154: this part of the comments confuses me. Is this a typo? Did something accidentally get deleted?

server/src/project/_utils/transform-milestones.ts
```diff
@@ -146,6 +146,21 @@ function applyDisplayDate(milestone, project) {
...
+  if((milestone.statuscode === 2) || (milestone.statuscode === 717170000)) {
+    if ((milestone._dcp_milestone_value === '8e3beec4-dad0-e711-8116-1458d04e2fb8') || (milestone._dcp_milestone_value === 'a43beec4-dad0-e711-8116-1458d04e2fb8')) { 
+      if((milestone.dcp_reviewmeetingdate) && (Date.now() >= milestone.dcp_reviewmeetingdate.getTime())) {
+        milestone.display_date = milestone.dcp_reviewmeetingdate;
+      }
+    }
+  }
 
   return milestone;
 }
```
godfreyyeung, line 166: I'm not sure the purpose of DisplayDate2, but in the templates the DisplayDate is conditionally displayed only if DisplayDate2 exists. Do you need to add this new logic to the `applySecondDisplayDate()` function as well?

server/src/project/_utils/transform-milestones.ts
godfreyyeung: Oh, you are clarifying the statuscode. 

Can you add that in? 

Something like "Statuscode optionset: 
2 == completed
717170000 = "In progress"

server/src/project/_utils/transform-milestones.ts
```diff
@@ -146,6 +146,21 @@ function applyDisplayDate(milestone, project) {
   if (milestone._dcp_milestone_value === '780593bb-ecc2-e811-8156-1458d04d0698') milestone.display_date = milestone.dcp_actualenddate;
   if (milestone._dcp_milestone_value === '483beec4-dad0-e711-8116-1458d04e2fb8') milestone.display_date = milestone.dcp_actualenddate;
   if (milestone._dcp_milestone_value === '4a3beec4-dad0-e711-8116-1458d04e2fb8') milestone.display_date = milestone.dcp_actualenddate;
+  
+  /*
+  * The following is for task 13239
+  * The date displayed should be the CRM Review Meeting date for certain milestones if the milestone is "in progress" or "completed", and the meeting date is <= current date
```
godfreyyeung, line 152: would be clearer if you wrote "if the milestone _statuscode_ is "in progress" or "completed"

server/src/project/_utils/transform-milestones.ts
dhochbaum-dcp: No, I just am not sure that there could not be an instance where all of the other criteria are met, but the CRM returns a null value for the reviewmeetingdate.  So I added a check to make sure it was not null, to ensure we wouldn't overwrite the date we want to use with a null value.

server/src/project/_utils/transform-milestones.ts
```diff
@@ -146,6 +146,21 @@ function applyDisplayDate(milestone, project) {
...
+  * I have included a check to make sure that the field is not null to make sure we are not overwriting dates with nothing
+  * 2 is completed, 717170000 is in progress
+  * Application Reviewed at City Planning Commission Review Session, City Planning Commission Vote - Need to get the ids for the other three milestones
+  */
+  if((milestone.statuscode === 2) || (milestone.statuscode === 717170000)) {
+    if ((milestone._dcp_milestone_value === '8e3beec4-dad0-e711-8116-1458d04e2fb8') || (milestone._dcp_milestone_value === 'a43beec4-dad0-e711-8116-1458d04e2fb8')) {
```
godfreyyeung, line 158: Nit: I think from both an algorithm and conceptual standpoint, it would be more efficient if the check for the milestone was the outer "if" statement, since that is more specific than statuscode. I bet more often a milestone would be completed or in progress than specifically those two milestone values.

server/src/project/_utils/transform-milestones.ts
dhochbaum-dcp: It looks to me like DisplayDate2 is only used if you want to show a date range, like "August 23, 2019 - September 6, 2019".  Otherwise it looks like it uses DisplayDate.  I don't think any modifications are necessary here.

godfreyyeung (approved):

server/src/project/_utils/transform-milestones.ts
dhochbaum-dcp: I like it!  I've pushed an update for this.

server/src/project/_utils/transform-milestones.ts
dhochbaum-dcp: I've updated the comment to clarify.

server/src/project/_utils/transform-milestones.ts
godfreyyeung: Oh I meant can you clarify the comment?


Also.... even if there is already data in `milestone.display_date`,  how do we know it's not a misleading date? Since we want to make sure all dates displayed are "dcp_reviewmeetingdate", wouldn't it be better to assign "null" if `dcp_reviewmeetingdate` is null?

server/src/project/_utils/transform-milestones.ts
godfreyyeung: oh I just realized that for those two specific miltestones, display_date_2 will always be null

server/src/project/_utils/transform-milestones.ts
godfreyyeung: Thanks!

godfreyyeung (approved):
Thanks for the quick changes! Approved.. again x)

PR #1255, opened 2020-12-08
13316 read document locations

godfreyyeung:
### Summary
Reworks the API to read the document folder URL for Packages and Artifacts, instead of reconstructing it from parts. This stabilizes the API since we no longer need to mimic the folder pattern dictated by CRM source code, which has been a moving target.

Also better organizes methods into different services.

  - 4 Sharepoint methods from crm.service.ts --> sharepoint.service.ts*
  - 2 Package methods from document.service.ts --> package.service.ts*
  - 2 Artifact methods from document.service.ts --> artifact.service.ts*

`*` = new service

This is "strangling" out the document service slowly :) 

Fixes [AB#13316](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/13316)


server/src/package/package.module.ts
```diff
@@ -0,0 +1,19 @@
+import { Module } from '@nestjs/common';
+import { DocumentModule } from '../document/document.module';
+import { SharepointModule } from '../sharepoint/sharepoint.module';
+import { PackageService } from './package.service'
+
+@Module({
+  imports: [
+    DocumentModule,
```
godfreyyeung, line 8: delete this doc module import?

server/src/artifact/artifact.module.ts
```diff
@@ -0,0 +1,19 @@
+import { Module } from '@nestjs/common';
+import { DocumentModule } from '../document/document.module';
+import { SharepointModule } from '../sharepoint/sharepoint.module';
+import { ArtifactService } from './artifact.service'
+
+@Module({
+  imports: [
+    DocumentModule,
```
godfreyyeung, line 8: delete doc module import?

server/src/artifact/artifact.service.ts
```diff
@@ -1,78 +1,103 @@
...
-      const errorMessage = `Error loading documents for artifact ${dcp_name}. ${JSON.stringify(e)}`;
-      console.log(errorMessage);
+    if (dcp_artifactdocumentlocation) {
+      try {
+        // Here we assume that the "entity" segment of the dcp_artifactdocumentlocation URL
+        // will be dcp_artifacts, since it is a artifact document location.
+        // Also that the segment divides the absolute URL into two tokens.
+        // Example value for dcp_artifactdocumentlocation:
+        // https://nyco365.sharepoint.com/sites/dcppfsuat2/dcp_artifacts/P2016K0021 - Area Map - 1_77A5253923FBE911A9BC001DD8308EF1
+        const relativeUrl = dcp_artifactdocumentlocation.split('/dcp_artifacts/')[1];
```
allthesignals, line 80: Is there someting about this assumption that worries you? Seems reasonable to me.

allthesignals: What happens when there's not documentlocation? Is it a silent "failure"? Why call the method if the required information isn't there?

allthesignals (approved):

server/src/artifact/artifact.service.ts
godfreyyeung: Yeah, the unfortunate part is it's a silent failure. I don't know what we can do about that though. 

For packages, when that happens I log a warning, since usually by the time a package is public, it will have the document location initiated even if no files are uploaded. And if document location is null, that is an oddity. 

For artifacts, I can't determine the difference between no file uploaded intentionally, which happens often, and some other error causing the artifact document location to be null. So I can't log a warning without spamming  server logs unnecessarily

PR #1256, opened 2020-12-09
13246: zoning resolution dropdown filter

trbmcginnis:
**Summary**
This PR adds a new filter to the filters section for "Zoning Resolutions". It works similarly to the Action Type filter where users can search zoning resolution types from a dropdown, and then select multiple zoning resolutions so that the filtered projects have ANY of the zoning resolution types selected. 

**Addresses Task [AB#13246](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/13246)**

**Technical Explanation**
Each `dcp_projectaction` entity has a navigation property called `dcp_ZoningResolution`. The projectAction entity therefore has a field called `_dcp_zoningresolution_value` which lists the id for the associated zoning resolution. 

We load ALL possible zoning resolutions from CRM -- this functions as a lookup for zoning resolution `name`-to-`id`. The name field on the `dcp_zoningresolution` is called `dcp_zoningresolution`. We use these loaded zoning resolutions as the options in the filter dropdown. 

When a user selects a zoning resolution name from the dropdown, the id that is associated with the selected name appears in the queryparams. We then use this value to query for projectActions that have the field `_dcp_zoningresolution_value` equal to that id.

### DIFF
--- a/client/app/routes/show-geography.js
+++ b/client/app/routes/show-geography.js
@@ -1,7 +1,12 @@
 import Route from '@ember/routing/route';
 
 export default class ShowGeographyRoute extends Route {
-  setupController(controller) {
+  async model() {
+    return this.store.findAll('zoning-resolution');
+  }
+
+  async setupController(controller, model) {
+    super.setupController(controller, model);
     controller.fetchData.perform({ unloadAll: true });
   }
 }
--- a/server/src/project/geometry/geometry.service.ts
+++ b/server/src/project/geometry/geometry.service.ts
@@ -141,6 +141,9 @@ const QUERY_TEMPLATES = {
   'action-types': (queryParamValue) =>
     equalsAnyOf('dcp_name', queryParamValue, 'dcp_projectaction'),
 
+  'zoning-resolutions': (queryParamValue) =>
+    queryParamValue.map(value => `dcp_dcp_project_dcp_projectaction_project/any(o:o/_dcp_zoningresolution_value eq '${value}')`).join(' or '),
+
   boroughs: (queryParamValue) =>
     containsAnyOf('dcp_borough', coerceToNumber(mapInLookup(queryParamValue, BOROUGH_LOOKUP)), 'dcp_project'),
 
--- a/server/src/zoning-resolutions/zoning-resolutions.controller.ts
+++ b/server/src/zoning-resolutions/zoning-resolutions.controller.ts
@@ -1,0 +1,66 @@
+import {
+  Controller,
+  Get,
+  HttpException,
+  HttpStatus,
+} from '@nestjs/common';
+import { CrmService } from '../crm/crm.service';
+import { Serializer } from 'jsonapi-serializer';
+import { dasherize } from 'inflected';
+
+const ACTIVE_STATUSCODE = 1;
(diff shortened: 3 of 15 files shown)


trbmcginnis:
zoning_res_test

client/app/routes/show-geography.js
```diff
@@ -1,7 +1,13 @@
 import Route from '@ember/routing/route';
 
 export default class ShowGeographyRoute extends Route {
-  setupController(controller) {
+  async model() {
+    const zoningResolutions = await this.store.findAll('zoning-resolution');
+    return zoningResolutions;
```
allthesignals, line 6: Could just `return this.store.findAll('zoning-resolution');`

server/src/project/geometry/geometry.service.ts
```diff
@@ -141,6 +141,9 @@ const QUERY_TEMPLATES = {
   'action-types': (queryParamValue) =>
     equalsAnyOf('dcp_name', queryParamValue, 'dcp_projectaction'),
 
+  'zoning-resolutions': (queryParamValue) =>
+    queryParamValue.map(value => `dcp_dcp_project_dcp_projectaction_project/any(o:o/_dcp_zoningresolution_value eq '${value}')`).join(' or '),
```
allthesignals, line 145: I know it's not obvious but the query syntax here is wrong. It should be XML syntax, not OData. So, this will throw errors for the map.

allthesignals: It might be ignoring it because there's a separate "ALLOWED FILTERS" array.

server/src/zoning-resolutions/zoning-resolutions.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+  'dcp_zoningresolution',
+];
+
+@Controller('zoning-resolutions')
+export class ZoningResolutionsController {
+  constructor(
+    private readonly crmService: CrmService,
+  ) {}
+
+  // Extract the raw Express instance and pass to the query method
```
allthesignals, line 24: Comment isn't correct :P

allthesignals (approved):
Wow! Fantastic job :) thanks for figuring this out.

server/src/project/geometry/geometry.service.ts
trbmcginnis: @allthesignals  can you explain more about what needs to change?

allthesignals: Well, look at the syntax for the other helpers:
image

It's fetchXML query syntax. In your code, however, you're using OData Web API syntax.

What needs to change:

1. You'd have to make this code use FetchXML syntax.
2. You'd have to make sure this is actually filtering the map (in your screenshot, it doesn't appear to be filtering the map at all).

I'm OK with this being a separate pull request, but I would remove this line with a TODO.

Let me know!

trbmcginnis: Okay yes I see what you're saying. I'm going to go ahead and merge this and start working on the fetchxml stuff right away, thanks!

PR #1260, opened 2020-12-14
13520 Guard against a null bbl value

godfreyyeung:
Some projects have null BBLs. Prevent the backend and frontend from acting on them.

Fixes [AB#13520](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/13520)

### DIFF
--- a/server/src/project/geometry/geometry.service.ts
+++ b/server/src/project/geometry/geometry.service.ts
@@ -283,10 +283,16 @@ export class GeometryService {
       .map(block => `${block.block.substring(1)}`);
   }
 
+  // Warning! Returns either null or an Object
   async getBblsGeometry(bbls = []) {
-    if (bbls === null || bbls.length === 0) return null;
-
+    // Sometimes Dynamics populates the BBLs array with one element: null.
+    // So we make sure to remove all null values.
     const normalizedBbls = bbls.filter(Boolean);
+
+    // The carto fetch fails if we try to construct a query from an empty array,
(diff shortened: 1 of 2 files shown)


server/src/project/geometry/geometry.service.ts
```diff
@@ -283,8 +283,11 @@ export class GeometryService {
       .map(block => `${block.block.substring(1)}`);
   }
 
+  // Warning! Returns either null or an Object
   async getBblsGeometry(bbls = []) {
-    if (bbls === null || bbls.length === 0) return null;
+    // Sometimes Dynamics populates the BBLs array with one element: null.
+    // The carto fetch fails if the one element is null.
+    if (bbls === null || bbls.length === 0 || bbls[0] === null) return null;
```
allthesignals, line 290: Thanks @godfreyyeung — I added a filter to the backend query to make sure it only returns not-null bbl.

My only problem with this approach is that a null value may be present in other parts of the array, at index 1 or 2, for example. Actually, one thing that'd help this is to just move line 292 above so that it filters out anything that isn't truthy (like nulls). Your call.

allthesignals (approved):

server/src/project/geometry/geometry.service.ts
godfreyyeung: Ahh, good suggestion. Will do!

godfreyyeung:
Made Matt's recommended changes, merging

PR #1262, opened 2020-12-22
13549: update cpc report links to pull from CRM field in projectaction entity

trbmcginnis:
### Summary
Update the CPC Report link that appears as a Ulurp number under the action item on a project page. The URL for this link will now pull from a new field in CRM. 

Test project: 2020M0334

### Task
Addresses [AB#13549](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/13549)

### Technical Explanation
There is now a field in CRM on the `dcp_projectaction` entity called `dcp_spabsoluteurl`. This field contains a link for the sharepoint CPC report documents folder. I updated the build-url helper so that the link in this field is passed into the build-url function instead of hardcoding the link that we had before.

### DIFF
--- a/client/app/helpers/build-url.js
+++ b/client/app/helpers/build-url.js
@@ -45,7 +45,7 @@ function bisweb(bbl) {
   return `http://a810-bisweb.nyc.gov/bisweb/PropertyBrowseByBBLServlet?allborough=${boro}&allblock=${block}&alllot=${lot}&go5=+GO+&requestid=0`;
 }
 
-function cpcReport(ulurp) {
+function cpcReport(ulurp, cpcUrl) {
   // for ulurp numbers that have a letter at the end of the 6 numbers to represent the version (ex. the "A" in C18005AZMX)
   // pull 6 numbers AND next character after the last number
   const ulurpNumberWithLetter = (ulurp.match(/(\d+)./g)[0]).toLowerCase();
@@ -53,10 +53,10 @@ function cpcReport(ulurp) {
 
   // the ulurp numbers with that extra letter have overall 11 characters compared to the usual 10
   if (ulurp.length > 10) {
-    return `http://www1.nyc.gov/assets/planning/download/pdf/about/cpc/${ulurpNumberWithLetter}.pdf`;
+    return `${cpcUrl}/${ulurpNumberWithLetter}.pdf`;
   }
 
-  return `http://www1.nyc.gov/assets/planning/download/pdf/about/cpc/${ulurpNumberWithoutLetter}.pdf`;
+  return `${cpcUrl}/${ulurpNumberWithoutLetter}.pdf`;
 }
 
 function acris(bbl) {
@@ -80,7 +80,7 @@ export function buildUrl([type, value, option]) {
   if (type === 'zoningResolution') return zoningResolution(value);
   if (type === 'zola') return zola(value);
   if (type === 'bisweb') return bisweb(value);
-  if (type === 'cpcReport') return cpcReport(value);
+  if (type === 'cpcReport') return cpcReport(value, option);
   if (type === 'acris') return acris(value);
   if (type === 'CommProfiles') return CommProfiles(value, option);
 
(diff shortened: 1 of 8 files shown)


client/app/helpers/build-url.js
```diff
@@ -45,18 +45,18 @@ function bisweb(bbl) {
   return `http://a810-bisweb.nyc.gov/bisweb/PropertyBrowseByBBLServlet?allborough=${boro}&allblock=${block}&alllot=${lot}&go5=+GO+&requestid=0`;
 }
 
-function cpcReport(ulurp) {
+function cpcReport(ulurp, cpcUrl) {
```
godfreyyeung, line 48: love love love the good param name

client/app/templates/show-project.hbs
```diff
@@ -117,7 +117,7 @@
                             content='City Planning Commision Report'
                             tagName='span'
                           }}
-                            <a href={{build-url "cpcReport" action.dcpUlurpnumber}} target="_blank">
+                            <a href={{if (and (not-eq action.statuscode ACTION_STATUSCODE_ACTIVE) action.dcpSpabsoluteurl) (build-url "cpcReport" action.dcpUlurpnumber action.dcpSpabsoluteurl)}} target="_blank">
```
godfreyyeung, line 120: Hmm, this might be better if the if statement is outside the `<a>`. It would be a bit more readable imo, and also here it's not clear what happens if the condition is false. Would the link just go to nowhere?

server/queries/projects/show.sql
```diff
@@ -50,7 +50,8 @@ SELECT
       'dcp_prefix', a.dcp_prefix,
       'statuscode', a.statuscode,
       'dcp_ccresolutionnumber', a.dcp_ccresolutionnumber,
-      'dcp_zoningresolution', z.dcp_zoningresolution
+      'dcp_zoningresolution', z.dcp_zoningresolution,
+      'dcp_spabsoluteurl', a.dcp_spabsoluteurl
```
godfreyyeung, line 54: these one-letter variable names 🙈  zap search can def use maintenance haha

server/src/assignment/assignment.service.ts
```diff
@@ -234,7 +234,7 @@ function generateAssignmentsQueryObject(contact) {
     $expand: `
       dcp_dcp_project_dcp_communityboarddisposition_project($filter=${DISPOSITIONS_FILTER}),
       dcp_dcp_project_dcp_projectmilestone_project($filter=${MILESTONES_FILTER};$select=dcp_milestone,dcp_name,dcp_plannedstartdate,dcp_plannedcompletiondate,dcp_actualstartdate,dcp_actualenddate,statuscode,dcp_milestonesequence,dcp_remainingplanneddayscalculated,dcp_remainingplanneddays,dcp_goalduration,dcp_actualdurationasoftoday,_dcp_milestone_value,_dcp_milestoneoutcome_value),
-      dcp_dcp_project_dcp_projectaction_project($select=_dcp_action_value,dcp_name,statuscode,statecode,dcp_ulurpnumber,_dcp_zoningresolution_value,dcp_ccresolutionnumber),
+      dcp_dcp_project_dcp_projectaction_project($select=_dcp_action_value,dcp_name,statuscode,statecode,dcp_ulurpnumber,_dcp_zoningresolution_value,dcp_ccresolutionnumber,dcp_spabsoluteurl),
```
godfreyyeung, line 237: oh gosh... so many places to include the expand 😵 💫  I wonder why we need it here

godfreyyeung (approved):
Primarily just have one concern about the `if` statement within the `<a href="">`.
Thanks Taylor! You're killing these. So many little details.

server/queries/projects/show.sql
trbmcginnis: lol agreed!

client/app/templates/show-project.hbs
trbmcginnis: that's a good point! I'll make those changes

PR #1269, opened 2021-01-08
make getSharepointFolderFiles flex; recurse file/folders

allthesignals:
This change modifies the getSharepointFolderFiles method to make it more flexible, and to make it include 3 levels deep of nested subfolders. It then unnests and flattens files found.

image

### DIFF
--- a/server/src/sharepoint/sharepoint.service.ts
+++ b/server/src/sharepoint/sharepoint.service.ts
@@ -6,6 +6,21 @@ import {
 import * as Request from 'request';
 import { ConfigService } from '../config/config.service';
 
+// {
+//  Files: []
+//  Folders: []
+// }
+
+function unnest(folders) {
+  return folders
+    .map(folder => {
+      return [...folder['Files'], ...unnest(folder['Folders'])];
+    })
+    .reduce((acc, curr) => {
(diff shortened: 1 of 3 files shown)


server/src/sharepoint/sharepoint.service.ts
```diff
@@ -6,6 +6,21 @@ import {
 import * as Request from 'request';
 import { ConfigService } from '../config/config.service';
 
+// {
+//  Files: []
+//  Folders: []
+// }
+
+function unnest(folders) {
```
godfreyyeung, line 14: Can you clarify and document the structure of the input and outputs from this function?
This is my _guess_ from reading the code, but just want to double check with you:
```
/**
 * @param { 
       { 
          Files: [],
          Folders: [],
       }[]
     } folders - array of "fileFolder" objects. "fileFolder" objects have a list of files and a recursive list of fileFolder objects.

 * returns {Document Objects []}
 */
 ```

godfreyyeung (approved):
I think this makes sense. So we only go 3 or 4 folders deep?

allthesignals:
> I think this makes sense. So we only go 3 or 4 folders deep? 

Yup

PR #1272, opened 2021-01-12
Remove use of Folder and File properties in shpt service to match UAT2 responses

godfreyyeung:
According to @allthesignals , production Sharepoint may deliver a different response format for files than UAT2.

Production now returns a response with `Folders` and `Files` properties, but UAT2 does not. Instead it holds an array of all package documents in a `value`  property.

Related to AB#769

@ksinghprod


godfreyyeung:
Closing this because UAT2 seems to have synced with Prod, and this is no longer an issue.

PR #1281, opened 2021-01-22
799 Stream CPC report links through ZAP API

godfreyyeung:
### Summary 
Previously, the CPC report links directly went to Sharepoint. For example
image
Would link to a URL like: "https://nyco365.sharepoint.com/sites/dcppfsuat2/dcp_projectaction/ZM - Zoning Map Amendment _D6297F65E0K5EB11V812001DD8Z0A6E2"

This is not going to work, because users would need to log in to sharepoint. 

This PR reworks the links so they follow the same patterns as for Packages and Artifacts under the Public Documents Section

### Details
This PR is bulky because it required componentizing the frontend to introduce some getters, avoiding complex template logic. The getters are needed to generate the correct "server relative" url to send to the API. 

Unlike documents for Packages and Artifacts, we do not enhance the Action model with our own custom Documents property. Thus the Action records don't already have a "serverRelativeUrl" property. We create a `serverRelativeURl` for individual actions in the `ShowProject::ActionsList::UlurpNumberLink` component.  I took some url-building code from the `build-url` ember helper to generate correct ulurp number-based file names.

Additionally...
  - I renamed the  `documentService.getDocument()` method since it no longer just handles Packages and Artifacts, but Actions as well.
  - I realized that we can determine what Entity the record id (derived from the document folder name) by just looking at the `/<entity_type>/`  segment of the URLs. So I removed the "trial-and-error" method of trying to query one entity after another for a certain record ID. I replaced it with simple if statements: `if (entityType === "dcp_package")`, `if (entityType === "dcp_artifact")`, `if (entityType === "dcp_projectaction")`
  
 
Fixes AB#799


server/src/document/document.service.ts
```diff
@@ -119,11 +138,30 @@ export class DocumentService {
...
+
+          throw new HttpException({
+            "code": "NO_DOCUMENT_ACCESS",
+            "detail": errorMessage,
+          }, HttpStatus.NOT_FOUND);
+        }
+      }
+
+      // similarly secure project action documents
+      if (entityType === "dcp_projectaction") {
```
allthesignals, line 153: I mean, I guess this is fine, but I don't understand why we keep adding more `if`s into a method, making it longer, making it harder to understand. Make more methods. Split stuff out. Etc.

client/app/components/show-project/actions-list/ulurp-number-link.js
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * @param {Action Object} action
+ */
+export default class UlurpNumberLinkComponent extends Component {
+  host = ENV.host;
+
+  ACTION_STATUSCODE_ACTIVE = 1;
+
+  get serverRelativeUrl () {
+    const serverRelativeFolderUrl = this.action.dcpSpabsoluteurl.replace('https://nyco365.sharepoint.com', '');
```
allthesignals, line 13: do UAT2 and prod have the same sharepoint folders?

client/app/components/show-project/actions-list/ulurp-number-link.js
```diff
@@ -0,0 +1,28 @@
+import Component from '@ember/component';
+import ENV from 'labs-zap-search/config/environment';
+
+/**
+ * @param {Action Object} action
+ */
+export default class UlurpNumberLinkComponent extends Component {
```
allthesignals, line 7: I see, you pulled the method from a helper. Why not just make a new helper? Why a component?

allthesignals: So now when peple see "build-url" they have to know that ulurp number link is an exception and use a component version of the same logic instead? I don't understand this decision.

client/app/templates/components/show-project/actions-list/ulurp-number-link.hbs
```diff
@@ -0,0 +1,29 @@
...
+        {{/tool-tipster}}
+      {{else}}
+        {{@action.dcpUlurpnumber}}
+      {{/if}}
+  {{else}}
+    {{@action.dcpUlurpnumber}}
+  {{/if}}
+{{else}}
+  No ULURP Number
+{{/if}}
```
allthesignals, line 29: Does the no newline thing not bother you?

client/app/components/show-project/actions-list/ulurp-number-link.js
allthesignals: You can also just export the function from the helper and import it here?

client/app/templates/components/show-project/actions-list/ulurp-number-link.hbs
```diff
@@ -0,0 +1,29 @@
...
+      @action.statuscode
+    )
+  }}  
+      {{#if (and (not-eq @action.statuscode this.ACTION_STATUSCODE_ACTIVE) @action.dcpSpabsoluteurl)}}
+        {{#tool-tipster
+          content='City Planning Commision Report'
+          tagName='span'
+        }}
+          <a
+            href={{concat this.host '/document' this.serverRelativeUrl}}
```
allthesignals, line 14: Why not just use the `build-url` helper here? Instead of pulling it out into another component?

server/src/document/document.controller.ts
```diff
@@ -70,9 +70,10 @@ export class DocumentController {
     }
   }
 
+  
   @Get('/*')
-  async read(@Param() path, @Res() res) {
-    const stream = await this.documentService.getPublicPackageOrArtifactDocument(path);
+  async read(@Param() sharepointRelativePath, @Res() res) {
+    const stream = await this.documentService.getDocument(sharepointRelativePath[0]);
```
allthesignals, line 76: Okay. I like what you did in documentService, where you used array destructuring to add clarity to what was being destructured. But you went backwards here, using index notation.

client/app/templates/show-project.hbs
```diff
@@ -88,71 +88,9 @@
...
-                    {{#if action.dcpCcresolutionnumber}}
-                      <p class="text-small no-margin">City Council Resolution: {{action.dcpCcresolutionnumber}}</p>
-                    {{/if}}
-                  </div>
-                </li>
-              {{/each}}
-            </ul>
+            <ShowProject::ActionsList
+              @projectActions={{model.actions}}
+            />
```
trbmcginnis, line 93: This being componentized makes it so much cleaner, thanks!

server/src/document/document.controller.ts
trbmcginnis, line 76: Just curious about the sharepoint process since I'm a bit unfamiliar -- the `sharepointRelativePath` comes through as an array?

client/app/templates/components/show-project/actions-list/ulurp-number-link.hbs
```diff
@@ -0,0 +1,29 @@
+{{#if @action.dcpUlurpnumber}}
```
allthesignals, line 1: Why is this its own component and not other parts of markup in actions-list? What if you rolled this back into actions-list, and looked at it more broadly: you loop over actions, there's some delegation behavior with `if`s, and use build-url consistently.

What you're proposing here looks like an ad hoc decision — it "feels" like it's "a lot" of markup, with some state (through relativeUrl), which I've pointed out seems unnecessary, so therefore, it should be its own component. Why? Why not make dcpZoningresolution and dcpCcresolutionnumber into their own components? I know they're tiny, but at least it'd make sense the second time going into it.

allthesignals: Even more confusing is the fact that the `if` block is evaluated in the component for the situation itself, rather than where every other type is evaluated, inside the actions-list. I would look at this again and try to imagine how someone who has never seen the code before would think this through.

server/src/document/document.service.ts
allthesignals: I know the code is sloppy almost everywhere else, but we shouldn't add to it.

allthesignals: https://sandimetz.com/99bottles

client/app/templates/components/show-project/actions-list.hbs
```diff
@@ -0,0 +1,45 @@
...
+            </a>
+          </p>
+        {{/if}}
+        {{#if projectAction.dcpCcresolutionnumber}}
+          <p class="text-small no-margin">City Council Resolution: {{projectAction.dcpCcresolutionnumber}}</p>
+        {{/if}}
+      </div>
+    </li>
+  {{/each}}
+</ul>
```
allthesignals, line 45: Can we add linting or something. Even Github's default UI doesn't like it.

client/app/components/show-project/actions-list/ulurp-number-link.js
```diff
@@ -0,0 +1,28 @@
...
+    // pull 6 numbers AND next character after the last number
+    const ulurpNumberWithLetter = (ulurpNumber.match(/(\d+)./g)[0]).toLowerCase();
+    const ulurpNumberWithoutLetter = ulurpNumber.match(/\d+/g)[0]; // pull 100149 from C100149ZSM
+
+    // the ulurp numbers with that extra letter have overall 11 characters compared to the usual 10
+    if (ulurpNumber.length > 10) {
+      return `${serverRelativeFolderUrl}/${ulurpNumberWithLetter}.pdf`;
+    }
+
+    return `${serverRelativeFolderUrl}/${ulurpNumberWithoutLetter}.pdf`;
```
allthesignals, line 26: This was hard for me to grock but understand it's a weird formatting. I don't know how wrote this originally, and I realize you're moving code over from something else, but I would consider ways to improve this code.

allthesignals (changes requested):
> This PR is bulky because it required componentizing the frontend to introduce some getters, avoiding complex template logic.

With respect, I find this change actually adds complex template logic.

Do we know for sure that these urls can't just be computeds on the action model? What can go on the model?

server/src/document/document.controller.ts
godfreyyeung: @allthesignals  I tried destructing here, like `@Param() [sharepointRElativePath]`, but there's some funky stuff going on with TypeScript decorators not liking destructuring. I think the `@Param()` decorator expects something atomic passed in. I looked into ways to work around it but they seemed to add more complexity. Do you know a way?

godfreyyeung: Yeah, the `@Param` decorator returns an array. I thought it was as bit weird, too. I think it's an abstraction to handle some more complex URLs?

client/app/components/show-project/actions-list/ulurp-number-link.js
```diff
@@ -0,0 +1,28 @@
+import Component from '@ember/component';
+import ENV from 'labs-zap-search/config/environment';
+
+/**
+ * @param {Action Object} action
+ */
+export default class UlurpNumberLinkComponent extends Component {
+  host = ENV.host;
+
+  ACTION_STATUSCODE_ACTIVE = 1;
```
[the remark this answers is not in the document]
godfreyyeung: @allthesignals  I need to access this in the template though. Do you think it's better to have redundancy?
like 
```
ACTION_STATUSCODE_ACTIVE = 1;

export default class UlurpNumberLinkComponent extends Component {
    ACTION_STATUSCODE_ACTIVE = ACTION_STATUSCODE_ACTIVE;
```

client/app/components/show-project/actions-list/ulurp-number-link.js
godfreyyeung: They both have the "nyco364.sharepoint.com" domain, but differ in the "environment" segment of the document URL

client/app/templates/components/show-project/actions-list.hbs
godfreyyeung: Sorry, I'll fix

server/src/document/document.controller.ts
allthesignals: You don't need to destructure in the method signature. I'm not sure of a way. Have you found anything online?

client/app/components/show-project/actions-list/ulurp-number-link.js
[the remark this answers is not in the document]
allthesignals: The convention I usually follow with that is to define a camel-case variable:

`actionStatuscodeActive = ACTION_STATUSCODE_ACTIVE;`

I have wanted a way to lookup action codes through a helper. I know you built something similar in applicant portal. 

But it might not matter if you end up refactoring the approach by retaining the helper.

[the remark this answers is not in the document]
allthesignals: > The convention I usually follow with that is to define a camel-case variable:
> 
> `actionStatuscodeActive = ACTION_STATUSCODE_ACTIVE;`

Following up, my reasoning here is that ALL_CAPS usually indicates some kind of static, constant, easy to find, where camelCase is a class property.

server/src/document/document.service.ts
godfreyyeung: @allthesignals  I added a refactor commit that sets up separate GET endpoints for each of the entities. Can you take a look and let me know your thoughts?

client/app/components/show-project/actions-list/ulurp-number-link.js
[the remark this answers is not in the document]
godfreyyeung: Gotcha. Will update

server/src/document/document.controller.ts
godfreyyeung: Hmm okay. I could do something like 
```
const [ firstElem] = sharepointRelativePath;
const stream = await this.documentService.getDocument(firstElem);
```
though I don't think it's necessarily clearer. I'm not a fan of code golf, but in this case I think `[0]` is more to-the-point. Just a personal preference. I don't think it's a big deal either way, though.

godfreyyeung: Most answers online point to more complex code, and just suggest destructuring in the body to keep things simple.

server/src/document/document.service.ts
```diff
@@ -69,22 +56,60 @@ export class DocumentService {
...
+    private readonly sharepointService: SharepointService,
+  ){}
+  // NOTE: There are no guarantees that Filed Land Use documents will show up until we have secured Sprint 10 enhancements from the EAS team.
+  // This is because there have been instances that the dcpName portion for Filed LU is stripped of hyphens and spaces.
+  // After that, we can also consider updating the app to read document locations.
+  public stripDcpName(dcpName) {
+    return dcpName.replace(/'+/g, '').replace(/^\~|\#|\%|\&|\*|\{|\}|\\|\:|\<|\>|\?|\/|\||\"/g, '');
+  }
+
+  public constructFolderIdentifier(dcpName, recordId) {
```
allthesignals, line 105: Great!

server/src/document/document.service.ts
```diff
@@ -106,41 +131,57 @@ export class DocumentService {
...
-        if (!firstArtifact) {
-          const errorMessage = `No document access.`;
-          console.log(`Client attempted to retrieve document ${pathSegment}, but no associated public, submitted packages or artifacts were found.`);
-
-          throw new HttpException({
-            "code": "NO_DOCUMENT_ACCESS",
-            "detail": errorMessage,
-          }, HttpStatus.NOT_FOUND);
-        }
+        throwNoDocumentError(`Client attempted to retrieve document ${path}, but no associated public, submitted packages were found.`);
```
allthesignals, line 135: Nice!

allthesignals (approved):
@godfreyyeung perfect! Going to approve this on the condition that you agree with and make changes to the frontend stuff as well!

client/app/components/show-project/actions-list/ulurp-number-link.js
godfreyyeung: @allthesignals  I've refactored this to make the code a little closer to human-language, I think. Take a look and lmk if you think so!

client/app/components/show-project/actions-list/ulurp-number-link.js
godfreyyeung: > > This PR is bulky because it required componentizing the frontend to introduce some getters, avoiding complex template logic.
> 
> With respect, I find this change actually adds complex template logic.
> 
> Do we know for sure that these urls can't just be computeds on the action model? What can go on the model?

@allthesignals  I like the suggestion to try placing the URL-building (i.e. the serverRelativeUrl computed) on the Action model. I pushed a new commit to do that. It also uses the build-url helper again. I also removed the ulurpLinkNumber component.

Take a look and lmk what you think!

godfreyyeung: I origina

> So now when peple see "build-url" they have to know that ulurp number link is an exception and use a component version of the same logic instead? I don't understand this decision.

I originally broke out the `cpcReport` case from build URL because it was no longer correct. The buildUrl cpcReport worked off of absolute URL, but that was the cause of the bug (links were linking directly to sharepoint). I think my initial reaction was that buildUrl is a legacy component I wanted to stay away from.

Also buildUrl has a long list of "if" conditions, which I feel like is what we want to eventually move away from.

client/app/templates/components/show-project/actions-list/ulurp-number-link.hbs
godfreyyeung: I want to shift as much complex logic to the component/model/helper files as much as possible

client/app/templates/components/show-project/actions-list/ulurp-number-link.hbs
godfreyyeung: >  and looked at it more broadly: you loop over actions, there's some delegation behavior with ifs, and use build-url consistently.
I don't follow what you mean here by "delegation behavior with ifs..." 

> it "feels" like it's "a lot" of markup, with some state (through relativeUrl), which I've pointed out seems unnecessary, so therefore, it should be its own component. Why? 
Good observation. I think that was part of it. I was following our discussions from building long forms in Applicant Portal -- we observed a ton of unwieldy, lengthy templates, and thought it would be better if we could break these up into smaller components so they're more manageable and testable. 
However I can also see your point that this is "componentizing just because".

godfreyyeung: @allthesignals 
> Even more confusing is the fact that the if block is evaluated in the component for the situation itself, 

 Sorry I don't follow this 🤔

client/app/components/show-project/actions-list/ulurp-number-link.js
godfreyyeung: I've refactored, hoping it'll make it clearer!

godfreyyeung:
> @godfreyyeung perfect! Going to approve this on the condition that you agree with and make changes to the frontend stuff as well!

@allthesignals Many thanks for your thoughtful feedback and constructive criticism! Yes I've made frontend changes, too

server/src/document/document.controller.ts
allthesignals: @godfreyyeung uh, code golf? What is that? Yeah, if you say "firstElem," it doesn't tell you anything about what it actually is. I'm mostly just looking for a semantic naming for the thing. Your call.

client/app/templates/components/show-project/actions-list/ulurp-number-link.hbs
allthesignals: The template looks like this:

```
if (x) do this
if (y) do this
runSomethingThatEvaluates(z)
if (a) do this
```
rather than

```
if (x) do this
if (y) do this
if (z) do runSomething()
if (a) do this
```

It doesn't follow the pattern suggested in its context.

allthesignals: > I don't follow what you mean here by "delegation behavior with ifs..."

Sorry, I just mean that it renders something based on a conditional. It delegates the responsibility of rendering markup based on the conditional.

client/app/components/show-project/actions-list/ulurp-number-link.js
allthesignals: Great!

client/app/templates/components/show-project/actions-list/ulurp-number-link.hbs
godfreyyeung: Oh I see, so with both commentsyou're saying  it reads better when every "if" is visible at the parent template. It'll be clearer to readers that the pattern applies for each of those displayed properties of projectAction. 

Thanks!

server/src/document/document.controller.ts
godfreyyeung: Code golf is when programmers try to shorten code as much as possible, often to detriment of readability. Gotcha. Tbh I don't know why in this case sharepointRelativePath returns an array. The Nest docs say the `@Param` decorator should return an object?

PR #1282, opened 2021-01-25
Filter folders by correct status

allthesignals:
This PR adds filters to explicitly filter artifacts that we want people to see. Prevents Disabled artifacts from appearing.

### DIFF
--- a/server/src/project/project.service.ts
+++ b/server/src/project/project.service.ts
@@ -132,6 +132,12 @@ const ARTIFACT_VISIBILITY = {
   GENERAL_PUBLIC: 717170003,
 }
 
+const ARTIFACT_STATUSCODE = {
+  ACTIVE: 1,
+  INACTIVE: 2,
+  SUBMITTED: 717170000,
+};
+
 // configure received params, provide procedures for generating queries.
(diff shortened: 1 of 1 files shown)


godfreyyeung (commented):
It makes sense to me conceptually. However I think I avoided adding these conditions because they weren't part of the original user requirement, and I wasn't expert enough in the planning domain to know if it actually made a positive difference.

The original user story only says to filter artifacts by their visibility.

Original user story:
https://dcp-paperless.visualstudio.com/ZAP%20Portals/_workitems/edit/13582?src=WorkItemMention&src-action=artifact_link
image

allthesignals:
> It makes sense to me conceptually. However I think I avoided adding these conditions because they weren't part of the original user requirement, and I wasn't expert enough in the planning domain to know if it actually made a positive difference.
> 
> The original user story only says to filter artifacts by their visibility.
> 
> Original user story:
> https://dcp-paperless.visualstudio.com/ZAP%20Portals/_workitems/edit/13582?src=WorkItemMention&src-action=artifact_link
> image

Maybe it was accidentally omitted in the user story?

allthesignals:
Closing

PR #1283, opened 2021-01-25
Use system-provided relative URL; expand subfolders in documents

allthesignals:
Previously, I reversed some of @godfreyyeung's changes to infer the document sharepoint path from an assumed convention. However, because these are separate systems, downstream edits were not being synchronized, leading to failing queries.

Now, this uses the provided FULL path, for packages, which is then split into a relative path, and passed along.

Artifacts, as @godfreyyeung has discovered and designed around, expresses package paths a little differently, not providing them as true relative paths (with in the packages means removing the host from the url), but instead as the plain folder names.

This change handles these separately.

TODO: Major major refactors to package/artifact services.

### DIFF
--- a/server/src/sharepoint/sharepoint.service.ts
+++ b/server/src/sharepoint/sharepoint.service.ts
@@ -75,12 +75,15 @@ export class SharepointService {
     try {
       const { access_token } = await this.generateSharePointAccessToken();
       const SHAREPOINT_CRM_SITE = this.config.get('SHAREPOINT_CRM_SITE');
+      let [, relativeUrl] = folderIdentifier.split(SHAREPOINT_CRM_SITE);
 
-      // Escape apostrophes by duplicating any apostrophes.
-      // See https://sharepoint.stackexchange.com/a/165224
-      const formattedFolderIdentifier = folderIdentifier.replace(/'/g, "''");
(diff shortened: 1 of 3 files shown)


server/src/sharepoint/sharepoint.service.ts
```diff
@@ -78,9 +78,15 @@ export class SharepointService {
 
       // Escape apostrophes by duplicating any apostrophes.
```
godfreyyeung, line 79: delete this

trbmcginnis (approved):
Thanks!

server/src/sharepoint/sharepoint.service.ts
```diff
@@ -75,12 +75,15 @@ export class SharepointService {
...
       const SHAREPOINT_CRM_SITE = this.config.get('SHAREPOINT_CRM_SITE');
+      let [, relativeUrl] = folderIdentifier.split(SHAREPOINT_CRM_SITE);
 
-      // Escape apostrophes by duplicating any apostrophes.
-      // See https://sharepoint.stackexchange.com/a/165224
-      const formattedFolderIdentifier = folderIdentifier.replace(/'/g, "''");
-      const url = encodeURI(`https://nyco365.sharepoint.com/sites/${SHAREPOINT_CRM_SITE}/_api/web/GetFolderByServerRelativeUrl('/sites/${SHAREPOINT_CRM_SITE}/${formattedFolderIdentifier}')/${path}`);
-
+      // package related urls are provided as true relative urls — but the relative url for artifacts is not relative but
+      // only provided as full path
```
godfreyyeung, line 81: Did you mean "// package relative urls are provided as true relative urls — but the url for artifacts is not relative. It is provided as a full path" ?

Honestly still confused what's going on here, though

PR #1285, opened 2021-01-27
Provide debug only link-out to CRM

allthesignals:
This PR adds a query param to the "show project" route, `debug`, which is false by default.

This query param is used to show or hide a hyperlink that constructs a URL for the related CRM project page.

image

I'm add this to make it easier to get to the CRM information, which is the source of truth. This should make debugging a little easier.

### DIFF
--- a/client/app/models/project.js
+++ b/client/app/models/project.js
@@ -55,7 +55,7 @@ export default class ProjectModel extends Model {
 
   @attr('boolean') dcpFemafloodzonea;
 
-@attr('boolean') dcpFemafloodzoneshadedx;
+  @attr('boolean') dcpFemafloodzoneshadedx;
 
   @attr('boolean') dcpSisubdivision;
 
@@ -64,6 +64,8 @@ export default class ProjectModel extends Model {
   @attr('string') dcpProjectbrief;
 
   @attr('string') dcpProjectname;
+
+  @attr('string') dcpProjectid;
 
   @attr('string', { defaultValue: '' }) dcpPublicstatus;
 
--- a/client/app/templates/show-project.hbs
+++ b/client/app/templates/show-project.hbs
@@ -8,8 +8,13 @@
           {{else}}
             Project {{model.dcpName}}
           {{/if}}
-        </h1>
 
+          {{#if this.debug}}
+            <a href="{{link-to-crm 'dcp_project' model.dcpProjectid}}" target="_blank">
+              {{fa-icon 'external-link-alt'}}
+            </a>
+          {{/if}}
+        </h1>
         <div class="grid-x">
           <div class="cell medium-auto">
             <p class="lead">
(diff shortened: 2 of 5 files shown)


client/app/models/project.js
```diff
@@ -65,6 +65,8 @@ export default class ProjectModel extends Model {
 
   @attr('string') dcpProjectname;
 
+  @attr('string') dcpProjectid;
```
godfreyyeung, line 68: In terms of process, these should be in a separate commit, instead of hidden in a commit indicating something completely different

client/app/templates/show-project.hbs
```diff
@@ -8,8 +8,13 @@
           {{else}}
             Project {{model.dcpName}}
           {{/if}}
-        </h1>
 
+          {{#if this.debug}}
+            <a href="https://nycdcppfs.crm9.dynamics.com/main.aspx?etn=dcp_project&id={{{model.dcpProjectid}}}&pagetype=entityrecord" target="_blank">
```
godfreyyeung, line 13: This should be dependent on environment. We're usually developing locally, and working with data on UAT2. UAT2 data can be pretty different from prod data

godfreyyeung (changes requested):

client/app/models/project.js
allthesignals: > In terms of process, these should be in a separate commit, instead of hidden in a commit indicating something completely different

I'm sorry — what do you mean by this? It's already being shipped with the response, and used in a template. How does the commit "indicate something completely different"?

client/app/templates/show-project.hbs
allthesignals: > This should be dependent on environment. We're ~usually~ always developing locally, and working with data on UAT2. UAT2 data can be pretty different from prod data

Sure, I can add that.

client/app/models/project.js
allthesignals: Confused about why you think this is some kind of process violation — it's a change that's necessary for this commit to function at all.

godfreyyeung: Oh sorry, the green over the dcpFemafloodzoneshadedx made me think multiple attributes were being added. 
In this case, why can't `project.id` or `project.dcpName`  be used, instead of dcpProjectid?

allthesignals: > Oh sorry, the green over the dcpFemafloodzoneshadedx made me think multiple attributes were being added.
> In this case, why can't `project.id` or `project.dcpName` be used, instead of dcpProjectid?

Because this model was set up to identify projects by their dcpName values, not the regular CRM ids.

We can't use dcpName because you can't link out to CRM projects with the dcpName value. It requires the GUID.

godfreyyeung: Ah okay, gotcha, thank you

godfreyyeung (approved):

trbmcginnis (approved):

PR #1286, opened 2021-01-27
Fix artifact naming

allthesignals:
Review after commits from #1285.

This change defaults to the artifact name if the formatting fails (due to the absence of a dash).

PR #1287, opened 2021-01-27
769 alphabetize doc list

allthesignals:
Please review commits after #1286 

Leverages the sort-by helper to alphabetize docs by their serverRelativeurl. Sorting by relative url is intended to capture some of the nested folder logic from in Sharepoint.

### DIFF
--- a/client/app/components/show-project/public-documents/artifact-documents.js
+++ b/client/app/components/show-project/public-documents/artifact-documents.js
@@ -10,7 +10,7 @@ export default class ArtifactDocumentsComponent extends Component {
   @tracked showArtifactDocuments = false;
 
   get formattedArtifactName() {
-    return this.artifact.dcpName.split(' - ')[1];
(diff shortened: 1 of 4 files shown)


trbmcginnis (approved):
Looks great! -- I know you mention that:

"Sorting by relative url is intended to capture some of the nested folder logic from in Sharepoint." 

^^ would you mind providing a bit more explanation on how this works?

allthesignals:
> Looks great! -- I know you mention that:
> 
> "Sorting by relative url is intended to capture some of the nested folder logic from in Sharepoint."
> 
> ^^ would you mind providing a bit more explanation on how this works?

Yup, the relative URL includes the subfolder names. So if we alphabetize by the URL, it'll also sort by those subfolders, listing stuff together a little bit. 

Definitely not an ideal solution but a good first step. We need to work with Aline to figure out best path forward!

PR #1289, opened 2021-01-27
Handle situations where there is only one null filter value

allthesignals:
Before, it was generating markup for the null, but still included syntax for non-null conditionals (“in”). This was throwing a syntax error CRM-side.

Now, it makes the non-null filter list blank if there are no non-null filters to apply.

Closes ADO#630.


trbmcginnis (approved):
Thanks!

PR #1290, opened 2021-01-29
788 Convert displayDate(2) to computeds on frontend Milestone model

godfreyyeung:
### Summary
This PR removes backend code that added the `displayDate` and `displayDate2` properties to milestones sent to the frontend. It sets up `displayDate` and `displayDate2` as computed properties on the frontend Milestone model, instead.

#### Why? 
The primary reason is to clean up code -- move things to where they belong -- to be more readable and maintainable.
1) Code that computes _derived_ properties for display belong in the frontend. This makes the backend unopinionated in how it delivers data.

    `displayDate` and `displayDate2` are not real properties in the CRM milestone table. [They are computed off `dcpActualstartdate`, `dcpActualenddate`, and `dcpReviewmeetingdate`](https://github.com/NYCPlanning/labs-zap-search/pull/1290/files#diff-3cb281dcf0b415503e45ea663c63a42115e63a34eea5df328a7a7dad7af349eaL127). Perfect for frontend computed properties.

2) The logic that calculates `displayDate`/`displayDate2` (now moved to the frontend) ultimately affected whether dates are shown under milestones on the project profile. That logic determines on a [milestone-by-milestone basis whether `displayDate`/`displayDate2`  receives a value or `null`](https://github.com/NYCPlanning/labs-zap-search/pull/1290/files#diff-3cb281dcf0b415503e45ea663c63a42115e63a34eea5df328a7a7dad7af349eaL127-L131). The [frontend templates rely on assignment of `null`](https://github.com/NYCPlanning/labs-zap-search/blob/develop/client/app/templates/components/project-milestone.hbs#L30) to determine if dates should appear under milestones. 

    Moreover, there is also other, miscellaneous frontend component code that determines if dates appear under milestones  (e.g. the [`tense()` CP]( https://github.com/NYCPlanning/labs-zap-search/blob/develop/client/app/components/project-milestone.js#L21) or [same-year checks](https://github.com/NYCPlanning/labs-zap-search/blob/develop/client/app/templates/components/project-milestone.hbs#L34 )) . It was hard to switch between backend and frontend portions to grok all logic that affects date display.

    _All logic that affects user interface display should be consolidated in the frontend, not spread out between a transformation function in the backend and templates in the frontend_

3) The logic that "determines on a milestone-by-milestone basis whether `displayDate`/`displayDate2`  receives a value or `null`" is not written well. There is an [`if` statement for each milestone, and the if statements match against raw milestone IDs](https://github.com/NYCPlanning/labs-zap-search/pull/1290/commits/c17fb5fd3bb9993a5557705c99188d58398bfbee#diff-3cb281dcf0b415503e45ea663c63a42115e63a34eea5df328a7a7dad7af349eaL55-L179). I've [rewritten it to be clearer using arrays and human-readable constants](https://github.com/NYCPlanning/labs-zap-search/pull/1290/files#diff-10eb090a45e6f64b690259ea6e32805e11e5fa6979be54fd88f1492b05b509dcR135-R146).

#### Other necessary technical changes
- Previously the backend  also sorted milestones immediately after assigning `displayDate(2)`. Since the sorting depends on the computed value of `displayDate(2)`, I had to also [move the displayDate-dependent milestone sorting to the frontend](https://github.com/NYCPlanning/labs-zap-search/pull/1290/files#diff-ab2f935a8a2e006b09a97f31181fb4b8352479e1ae3fabc3a46bd89d514cd22fR127-R141). 

#### Tasks/Bug Numbers
- This PR sets up foundation for future PRs fixing AB#788

  - Start with looking at the first commit, which is the main work of moving code from the backend. The second commit deletes `displayDate(2)` from mirage factories and test setup data, since in reality the backend no longer delivers it.
   - Also see comments I added below


client/app/models/milestone/constants.js
```diff
@@ -0,0 +1,105 @@
...
+  'to-review': TO_REVIEW_MILESTONES,
+  reviewed: REVIEWED_MILESTONES,
+  archive: ARCHIVE_MILESTONES,
+};
+
+export const REFERRAL_MILESTONEID_BY_ACRONYM_LOOKUP = {
+  CB: COMMUNITY_BOARD_REFERRAL,
+  BP: BOROUGH_PRESIDENT_REFERRAL,
+  BB: BOROUGH_BOARD_REFERRAL,
+};
```
godfreyyeung, line 111: Ok I see you

client/app/components/project-milestone.js
```diff
@@ -32,30 +32,4 @@ export default class ProjectMilestoneComponent extends Component {
 
     return 'future';
   }
-
-  @computed('tense', 'milestone.{displayDate,displayDate2}')
-  get timeRelativeToNow() {
```
godfreyyeung, line 37: This was not used anywhere, so deleting.

client/app/models/project.js
```diff
@@ -124,4 +125,19 @@ export default class ProjectModel extends Model {
...
+      if (!prev.displayDate)
+        return 1;
+      if (!next.displayDate)
+        return -1;
+      return prev.displayDate - next.displayDate;
+    }
+
+    return milestoneSequenceDifference;
+  })
+  sortedMilestones;
```
godfreyyeung, line 142: Previously, the backend sorted immediately after calculating displayDate/displayDate2. Now that those properties are on the frontend, we also sort on the frontend. Sorting on the frontend make more sense, too, because it is related to data display.

This is essentially a migration of the [`sortMilestones` function from the backend to frontend](https://github.com/NYCPlanning/labs-zap-search/pull/1290/files#diff-3cb281dcf0b415503e45ea663c63a42115e63a34eea5df328a7a7dad7af349eaL234-L241)

client/tests/acceptance/hearings-list-for-milestones-list-shows-up-correctly-test.js
```diff
@@ -425,11 +415,9 @@ module('Acceptance | hearings list for milestones list shows up correctly', func
           milestonename: 'Land Use Application Filed',
           dcpMilestone: '663beec4-dad0-e711-8116-1458d04e2fb8',
           outcome: null,
-          displayDate2: null,
-          displayDate: '2019-07-13T19:31:57.763Z',
           statuscode: 'Completed',
-          dcpActualenddate: null,
-          dcpActualstartdate: '2019-07-13T19:31:57.763Z',
+          dcpActualenddate: '2019-07-13T19:31:57.763Z',
+          dcpActualstartdate: null,
```
godfreyyeung, line 420: `Moving the displayDate calculation code to the frontend actually revealed this mistake in the factory setup`, since mirage no longer hardcodes displayDate(2) and hidden from the CP implementation.

client/tests/acceptance/hearings-list-for-milestones-list-shows-up-correctly-test.js
```diff
@@ -522,11 +504,9 @@ module('Acceptance | hearings list for milestones list shows up correctly', func
...
```
godfreyyeung, line 509: see comment above

allthesignals:
Excited to look at this! Love the links to design principles

client/app/components/project-milestone.js
allthesignals: ty

client/app/models/milestone.js
```diff
@@ -107,4 +129,88 @@ export default class MilestoneModel extends Model {
...
+    if ([
+      BOROUGH_BOARD_REFERRAL,
+      BOROUGH_PRESIDENT_REFERRAL,
+      COMMUNITY_BOARD_REFERRAL,
+      CITY_COUNCIL_REVIEW,
+      CPC_PUBLIC_MEETING_PUBLIC_HEARING,
+      EIS_DRAFT_SCOPE_REVIEW,
+      FEIS_SUBMITTED_AND_REVIEW,
+      MAYORAL_VOTE,
+    ].includes(this.dcpMilestone) && this.project.get('dcpPublicstatus') !== 717170000) {
```
allthesignals, line 144: Do you know if we can use native getter syntax for this? `this.project.dcpPublicstatus`. I know that it being a model relation complicates matters for asynchronous relationships.

allthesignals: Can `717170000` be a named constant somewhere? Even the top of the file?

allthesignals: I would even just define `dcpPublicstatus` at the top `const dcpPublicstatus = ...` just to reduce the cognitive overhead of the syntax (native vs classic getters)

client/app/models/milestone.js
```diff
@@ -107,4 +129,88 @@ export default class MilestoneModel extends Model {
...
+    ].includes(this.dcpMilestone)) {
+      if ((this.statuscode === 2) || (this.statuscode === 717170000)) {
+        if ((this.dcpReviewmeetingdate) && (Date.now() >= this.dcpReviewmeetingdate.getTime())) {
+          return this.dcpReviewmeetingdate;
+        }
+      }
+    }
+
+    return null;
+  }
```
allthesignals, line 199: Only alternative approach I can think of is to group things by whether their project `dcpPublicstatus` value is/isn't 717170000. Can you tell me if this computed is "order-sensitive"? For example, if you moved one of these conditional blocks around, would it behave differently? If so, it's sort of an implicit behavior, and could easily confuse someone later.

client/app/models/project.js
allthesignals: Great

allthesignals (approved):
Excellent, thorough, clear. Just a few stray questions.

client/app/models/milestone.js
godfreyyeung: Yeah Ember was complaining if I used the native getter syntax -- the error said it needs the Ember getter because of the property is relationship proxy. 

Yes good idea about the constant.  Will change.

godfreyyeung: Ahh okay, I'll get it at the top.

client/app/models/milestone.js
godfreyyeung: Oh wow, really good question and something I didn't even realize. Here it _is_ order sensitive but that is a mistake/bug during this refactor! In the original transform, each conditional block only assigned to a value, but did not immediately `return`. The last block here was run _in addition_ to previous blocks for certain milestones, in order to override the value assigned by previous blocks. It previously wasn't order sensitive.

However I changed it so each block `return`s, and that makes it order sensitive. Will change it! THANK YOU @allthesignals.

PR #1292, opened 2021-02-02
Replace MilestoneConstants service with simple ES6 exports

godfreyyeung:
### Summary
Previously, we use an Ember service to reference Milestone IDs through human-readable symbols. This wasn't ideal because in the `models/milestone.js` file, I want to create lists of milestones in model computed properties. The service made it more verbose and awkward to reference.

For example, with a service a list of milestone would look like this:
```
[
      this.milestoneConstants.BOROUGH_BOARD_REFERRAL, 
      this.milestoneConstants.BOROUGH_PRESIDENT_REFERRAL, 
      this.milestoneConstants.COMMUNITY_BOARD_REFERRAL, 
      this.milestoneConstants.CITY_COUNCIL_REVIEW, 
      ....
 ]
```
(Or, we destructure at the beginning of the CP, but that would lead to redundant code in the CP.)

Now, by importing the constants at the top of the file, we don't need to reference the milestone service for every milestone, and reduce verbosity.
```
[
      BOROUGH_BOARD_REFERRAL, 
      BOROUGH_PRESIDENT_REFERRAL, 
      COMMUNITY_BOARD_REFERRAL, 
      CITY_COUNCIL_REVIEW, 
      ....
 ]
```
Also, I don't think a service is the right place to hold constants. A service is more for things like persisting sessions or shopping carts. As Ember is moving towards using more native ES6 features, we can follow suit by using simple ES6 exports and imports.

## Related Tickets
This PR is setting up a foundation for future commits related to AB#788


allthesignals (approved):
Awesome! Thank you

PR #1293, opened 2021-02-02
Use GH actions for client-side testing

allthesignals:
### Summary

Although there wasn't anything "wrong" with CircleCI, I think that service should be replaced with GH actions because:
 - We pay for GH
 - CircleCI has been failing randomly a few times inexplicably

This change begins the process of switching CircleCI with GH by creating a new Github action that tests the client-side code.

### Technical Explanation

We don't pay CircleCI and it's failing inexplicably. I want to switch to GH actions to run our tests because we pay for it so we should use it instead of a lower-grade service.

New Github actions are stored in the folder /.github/workflows. This adds /.github/workflows/main.yml.

It seems like new actions such as this are rolled into a new PR "check" when specifying pull_requests in its trigger.

  e. Provide links and explanations for any new technical concepts, APIs and terms.
      Especially if you had to do some research yourself.

I had to research here and ask @SPTKL a few qeustions. Here's a quickstart on it: https://docs.github.com/en/actions/quickstart


godfreyyeung (approved):
Cool! Thank you!!

SPTKL:
github actions is free & unlimited for public repos!

PR #1295, opened 2021-02-03
Remove CircleCI configuration

allthesignals:
### Summary

Because this repo is switching to Github Actions for running tests, we should removed CircleCI configuration folders. I've also disabled builds on CircleCI's side.

We used GH actions now. See #1293.


trbmcginnis (approved):
Thanks!

PR #1296, opened 2021-02-03
Merge crm adapter

allthesignals:
### Summary

The app wasn't handling network errors correctly due to outdated code, causing app-wide crashes or failed requests. This fix addresses that problem by swapping out the old code with a modern library, and merging some redundant code found in other classes.

#### Tasks/Bug Numbers
 - Fixes AB#843 (https://dev.azure.com/NYCPlanning/ITD/_workitems/edit/843/)

### Technical Explanation

Over the past few weeks, ZAP Search was experiencing app-wide crashes due to unhandled exceptions. Although we thought the unhandled exceptions were, in fact, handled, old syntax was being used in our CRM adapters which was circumventing our modernized approach to exception handling. Specifically, our modules were using [request](https://www.npmjs.com/package/request), which is deprecated, and it was using callbacks for Promises, which were explicitly rejecting but not being caught by try/catch blocks higher up in the chain. This was wreaking havoc.

The fix was swapping out Request for [superagent](https://github.com/visionmedia/superagent), which has full support for async await syntax. Further, it is a much more concise syntax, removing a lot of the boilerplate found in the old Request-based module.

On top of swapping out the libraries, this PR also merges some redundant classes together. More specifically, it eliminates the use of the static util class `_utils/crm-web-api.ts`, it deletes the odata module, it combines some odata service behavior into the crm.service, and it updates all references to the deleted odata module to the crm service throughout the app.

Note that the odata.module.ts file exported a number of app-wide static utility functions. I went ahead and combined these into the crm folder inside crm.utilities.ts.

See the diff for crm.service:

image

This style of handling HTTP requests and promises can be found throughout. 

  c. Provide entrypoint of new solution, if different than (b).

crm/crm.service.ts includes the biggest changes.

The app uses crm.service for all CRM interactions instead of a combination of odata and crm services. Odata service has been deleted. 

  e. Provide links and explanations for any new technical concepts, APIs and terms.
      Especially if you had to do some research yourself.

This was helpful: https://jakearchibald.com/2017/await-vs-return-vs-return-await/

   List any ad-hoc, miscellaneous updates included

There are 3 "types" of changes throughout:

1. Moves static utility functions from an odata module class to a standalone crm utilities class.

image

2. Replaces usage of the deprecated Request library with modern/maintained SuperAgent

image

3. Delete the odata module/service and replace imports of it with the enhanced CRM service

image

A few things to note:

Throughout the old CRM module, headers were being handled on a per-method-basis, with no standardization. This changes that by creating a generic `defaultHeaders` method:

image
image

Additional, method-specific headers may be added as an argument.


server/src/crm/crm.utilities.ts
```diff
@@ -109,5 +98,79 @@ export function containsAnyOf(propertyName, strings = [], options?) {
...
+}
+
+
+const COMMUNITY_DISPLAY_TOKEN = '@OData.Community.Display.V1.FormattedValue';
+
+// CRM provides numeric codes for picklist types
+// for example, "yes" might appear as "1"
+// This function maps those values with appropriate labels
+// TODO: Make this return new objects instead of in-place reassignment
+export function overwriteCodesWithLabels(records, targetFields) {
```
allthesignals, line 136: This was copied from odata.module.ts

server/src/project/project.service.ts
```diff
@@ -317,9 +315,7 @@ export function transformProjectAttributes(project): any {
 @Injectable()
 export class ProjectService {
   constructor(
-    private readonly dynamicsWebApi: OdataService,
     private readonly config: ConfigService,
-    private readonly carto: CartoService,
```
allthesignals, line 322: Miscellaneous. Wasn't being used anywhere in this service.

server/src/crm/crm.service.ts
```diff
@@ -3,21 +3,39 @@ import {
...
+      // return the raw response if the response parsing fails
+      err.rawResponse = res.text || null;
+
+      // return the http status code if the response parsing fails
+      err.statusCode = res.status;
+    } finally {
+      fn(err, body);
+    }
+  });
+};
```
allthesignals, line 34: This code extends superagent's default JSON parser and applies the "dateParser" code to it. Note that this is the same behavior in the old, static crm-web-api code. Previously, this was called "_dateReviver". You can find this old method in the diff.

server/src/crm/crm.service.ts
```diff
@@ -66,340 +116,46 @@ export class CrmService {
...
-          return new Date(Date.UTC(parseInt(a[1]), parseInt(a[2]) - 1, parseInt(a[3]), parseInt(a[4]), parseInt(a[5]), s, ms));
-        }
-
-        const b = /^(\d{4})-(\d{2})-(\d{2})$/.exec(value);
-
-        if (b) {
-          return new Date(parseInt(b[1]), parseInt(b[2]) - 1, parseInt(b[3]), 0, 0, 0, 0);
-        }
+  // TODO: Remove object-driven query templating
+  serializeToQueryString(query: any) {
```
allthesignals, line 145: Migrated from the odata.service

server/src/crm/crm.service.ts
```diff
@@ -66,340 +116,46 @@ export class CrmService {
...
-   * @param      {string}  entity     the pluralized version of the entity
-   * @param      {string}  xmlQuery   query string to additionally filter, in XML
-   * @return     {string}
-   */
-  async getWithXMLQuery(entity: string, xmlQuery: string, ...options) {
-    const response = await this._get(`${entity}?fetchXml=${xmlQuery}`, ...options);
-    return response;
-  }
+  // TODO: Remove object-driven query templating
+  async queryFromObject(entity: string, query: any, ...options) {
```
allthesignals, line 138: Holdover from odata.service.

server/src/crm/crm.utilities.ts
```diff
@@ -109,5 +98,79 @@ export function containsAnyOf(propertyName, strings = [], options?) {
...
+      const b = /^(\d{4})-(\d{2})-(\d{2})$/.exec(value);
+
+      if (b) {
+        return new Date(parseInt(b[1]), parseInt(b[2]) - 1, parseInt(b[3]), 0, 0, 0, 0);
+      }
+    }
+  }
+
+  return value;
+}
```
allthesignals, line 127: This is used in crm.service and does the same thing as the old _dateReviver method from the old crm module.

### Why

This detects Date-like strings when parsing JSON. There is no Date-type in JSON. So, when parsing JSON, this tries to help by converting those to JavaScritp Date Objects. Upstream, we have code that depends on this, so I've kept the behavior.

server/src/crm/crm.service.ts
```diff
@@ -3,21 +3,39 @@ import {
...
+    } finally {
+      fn(err, body);
+    }
+  });
+};
 
 /**
  * This service is responsible for providing convenience
  * methods for interacting with CRM.
- * TODO: Leverage the oData service written by @allthesignals
```
godfreyyeung, line 15: thx!

server/src/contact/contact.service.ts
```diff
@@ -1,13 +1,18 @@
...
+import {
+  all,
+  comparisonOperator,
+} from '../crm/crm.utilities';
 
 @Injectable()
 export class ContactService {
   constructor(
-    private readonly odata: OdataService,
+    private readonly odata: CrmService,
```
godfreyyeung, line 15: Prefer to change this reference to `crmService: CrmService` ?

server/src/disposition/disposition.controller.ts
```diff
@@ -51,7 +51,7 @@ const { deserialize } = new Deserializer({
 @Controller('dispositions')
 export class DispositionController {
   constructor(
-    private readonly dynamicsWebApi:OdataService,
+    private readonly dynamicsWebApi:CrmService,
```
godfreyyeung, line 54: Nit: Can we change this to crmService too? I think the consistency will help w readability, and may help with future find and replace

server/src/contact/contact.service.ts
allthesignals: Yep! Ty

server/src/disposition/disposition.controller.ts
allthesignals: Yess good eye, thanks

server/src/crm/crm.utilities.ts
godfreyyeung: Thank you

server/src/crm/xml/xml.service.ts
```diff
@@ -2,7 +2,6 @@ import { Injectable } from '@nestjs/common';
 import { CrmService } from '../crm.service';
 import * as request from 'superagent'
 import { ConfigService } from '../../config/config.service';
-import { ADAL } from '../../_utils/adal';
```
godfreyyeung, line 5: Can we delete the xml.service.ts altogether?

server/src/crm/crm.service.ts
```diff
@@ -45,17 +63,49 @@ export class CrmService {
...
+    };
+  }
+
+  async get(entity: string, query: string) {
     try {
       const sanitizedQuery = query.replace(/^\s+|\s+$/g, '');
-      const response = await this._get(`${entity}?${sanitizedQuery}`, ...options);
+      const defaultHeaders = await this.generateDefaultHeaders();
+      const { body } = await superagent
+        .get(`${this.host}${entity}?${sanitizedQuery}`)
```
godfreyyeung, line 97: wow.... sooo much shorter ✨

server/src/crm/crm.service.ts
```diff
@@ -66,340 +116,46 @@ export class CrmService {
...
-    };
-
-    return new Promise((resolve, reject) => {
-      Request.get(options, (error, response, body) => {
-        const encoding = response.headers['content-encoding'];
-
-        if (!error && response.statusCode === 200) {
-          // If response is gzip, unzip first
 
-          const parseResponse = jsonText => {
```
godfreyyeung, line 167: 🤮 

Thank you sooo much

server/src/crm/crm.service.ts
```diff
@@ -66,340 +116,46 @@ export class CrmService {
...
-            parseError(body);
-
-          }
-        }
-        else resolve();
-      })
-    });
-  }
-
-  async _associate(relationshipName, entitySetName1, guid1, entitySetName2, guid2, headers) {
```
godfreyyeung, line 398: Looks like we don't use this in either Search or Portal API, so deleting is ok

server/src/crm/crm.service.ts
```diff
@@ -66,340 +116,46 @@ export class CrmService {
...
-            parseError(body);
-
-          }
-        }
-        else resolve(body);
-      })
-    });
-  }
-
-  async _sendDeleteRequest(query, headers) {
```
godfreyyeung, line 354: WARNING: If we eventually merge Search and Portal APIs, make sure this deletion doesn't take precedent. 

Applicant Portal API depends on this method:
https://github.com/NYCPlanning/labs-applicant-portal/blob/develop/server/src/crm/crm.service.ts#L410

godfreyyeung: Maybe we should preserve  this function and migrate it to superagent as well?

server/src/crm/crm.service.ts
```diff
@@ -66,340 +116,46 @@ export class CrmService {
...
-            });
-          } else {
-            parseError(body);
-          }
-        }
-      });
-    });
-  }
-
-  async _create(query, data, headers): Promise<any> {
```
godfreyyeung, line 224: 👍🏾  Looks like we don't use _create in either Search or Portal API, so deleting is ok

server/src/project/project.service.ts
```diff
@@ -378,7 +374,7 @@ export class ProjectService {
       'dcp_dcp_project_dcp_projectaddress_project($select=dcp_name)',
     ];
     
-    const { records: projects } = await this.dynamicsWebApi
+    const { records: projects } = await this.crmService
```
godfreyyeung, line 377: 👍🏾

godfreyyeung (approved):
Have a few change requests but approving:
1) delete xml.service.ts
2) Use `crmService: CrmService` throughout. There is some construction of `dynamicsWebApi: CrmService` or `oDate: CrmService`
3) Preserve the `delete()` and `_sendDeleteRequest()` methods in `crm.service.ts` and likewise migrate to superagent, to prevent loss of the endpoint in a future merge of Search and Portal APIs

Also small nit (sorry probably annoying) but PR is a bit chunky, and breaking out some of these changes into separate PRs would have helped with some clarity!

Lovely! Thank you for thorough description! ☺️

allthesignals:
> 1. delete xml.service.ts

So, it gets used in an awkward way by setting itself onto the imported crmService (see line 31). So, we can't delete it. In a later PR, I'll flag that for refactoring because it's confusing.

> 2. Use `crmService: CrmService` throughout. There is some construction of `dynamicsWebApi: CrmService` or `oDate: CrmService`

Agree, will do.

> 3. Preserve the `delete()` and `_sendDeleteRequest()` methods in `crm.service.ts` and likewise migrate to superagent, to prevent loss of the endpoint in a future merge of Search and Portal APIs

Happy to do this in a follow up PR to avoid further "chunkiness" ;) 

> Also small nit (sorry probably annoying) but PR is a bit chunky, and breaking out some of these changes into separate PRs would have helped with some clarity!

Will do!

server/src/crm/xml/xml.service.ts
allthesignals: nope =/

server/src/crm/crm.service.ts
allthesignals: will do in follow up

server/src/crm/crm.service.ts
allthesignals: will re-add in follow up but it's simpler for me to not create methods that don't have any direct usage in the app itself

allthesignals:
Follow up things:

1. Abolish use of the crm-web-api static utility file — see documents controller/service — use crm service
2. Re-add other HTTP methods/interactions with CRM API

godfreyyeung:
> > 1. delete xml.service.ts
> 
> So, it gets used in an awkward way by setting itself onto the imported crmService (see line 31). So, we can't delete it. In a later PR, I'll flag that for refactoring because it's confusing.
> 

Ahh, bummer. Okay thanks!

> > 1. Preserve the `delete()` and `_sendDeleteRequest()` methods in `crm.service.ts` and likewise migrate to superagent, to prevent loss of the endpoint in a future merge of Search and Portal APIs
> 
> Happy to do this in a follow up PR to avoid further "chunkiness" ;)
> 
> > Also small nit (sorry probably annoying) but PR is a bit chunky, and breaking out some of these changes into separate PRs would have helped with some clarity!
> 
> Will do!

Sounds awesome. Thank you!!! 🙏🏾  :)

PR #1299, opened 2021-02-08
788 Base milestone style on statuscode, not date fields

godfreyyeung:
### Summary
Previously, the milestones relied on date fields to determine styling -- e.g. whether or not the milestone is grayed out to indicate a "future" milestone. The PR changes the styling to be based on the `statuscode` property. This should bring some more stability to the milestones list behavior.

Additionally, this PR
  - a) removes some redundant conditional checks surrounding date logic
  - b) removes brittle CSS logic, which also depended on dates, for marking "now" on the milestone list
  - c) fixes formatting surrounding date logic
  - d) [fixes minor code errors surrounding date logic](https://github.com/NYCPlanning/labs-zap-search/pull/1299/commits/4fba94174659fcbdc967c09aa8b952cd2b13f661)
  - e) Fixes tests

#### Why move away from the date fields? 
1) Requirements say that the milestone `statuscode` should determine their styling. I.e. whether the milestone is grayed out, normal, or has a green glow to indicate "past", "present" or "future" respectively.
2) The date fields in CRM are not reliable. Sometimes they exist, sometimes not. Of course, sometimes they're in the future and sometimes in the past. Moreover different milestones depend on different date fields. This meant that before this PR there were many moving parts that determined the milestone state.  
(This is especially true for UAT2 data)

#### Tasks/Bug Numbers
 - Fixes AB#788

### Technical Explanation
Previously, the `project-milestone.js` component had a computed property `tense()`. This was the core of how milestone styles were derived. It would either return 'past', 'present' or 'future'. Style classes would be tacked onto the milestone elements depending on the return value. To calculate the return value, it inspected the given milestone's `displayDate` and `displayDate2` properties, and checked for presence and relation to the current time.  

However, displayDate and displayDate2 is assigned from either `dcpActualstartdate`, `dcpActualenddate`, or `dcpReviewmeetingdate`. It's hard to predict when these fields are or aren't defined, and sometimes they are not synced with the milestone `statuscode`. Moreoever, there are also multiple conditions for assignment, and assignment depends on the milestone as well. 

_If, for any reason, `displayDate` or `displayDate2` are `undefined` or `null`, then the styling of the milestone is wrong._ The milestone would show as grayed out, even though the status is 'In Progress' or 'Completed'.

Requirements from Special Projects indicates we can use a more predictable field, `statuscode`, to determine the styling. Specifically, AB#788 has indicated that when milestones have a statuscode of `In Progress` or `Completed`, it should NOT be grayed out even if there is no date present.

#### Other Items
- a) Made project-milestone template and component logic more focused:
  - [Moved the listing of milestone outcomes inside the "if displayDate && statuscode == Approved/Completed"  condition](https://github.com/NYCPlanning/labs-zap-search/pull/1299/commits/f8e7d10154427027984aa8fbf728130f386c69c7#diff-1ba465157c4ec2bc819a296718d0222252b0f8add4b1ae40b346e5b1ab4a150eL71-L73).  They should only show if a milestone is in progress or completed with a date (Aline said).
  -  Since the template already ensures only In Progress or Completed milestones will display a date, we [remove the check in the milestone `displayDate` CPs on the statuscode](https://github.com/NYCPlanning/labs-zap-search/pull/1299/commits/f8e7d10154427027984aa8fbf728130f386c69c7#diff-10eb090a45e6f64b690259ea6e32805e11e5fa6979be54fd88f1492b05b509dcL190).
 - b) [Removed the CSS for placing a "now" marker, because it relied on the unreliable `tense()` CP](https://github.com/NYCPlanning/labs-zap-search/pull/1299/commits/675521acf254bc90f84db6bc1a654f4b1bbe4741). [See comment below](https://github.com/NYCPlanning/labs-zap-search/pull/1299#discussion_r571795469). 
- e)  The changes revealed that some tests had set up mirage data wrong. For example, CB Review milestones should be already completed and have `dcpActualenddate`, if there are review vote results to show.  Fixed that data.


client/app/models/milestone.js
```diff
@@ -182,13 +182,15 @@ export default class MilestoneModel extends Model {
 
     if ([
       REVIEW_SESSION_CERTIFIED_REFERRED, // aka 'Application Reviewed at City Planning Commission Review Session'
-      CPC_PUBLIC_MEETING_VOTE, // aka 'City Planning Commission Vote'
-      CPC_PUBLIC_MEETING_PUBLIC_HEARING, // aka 'CPC Public Meeting - Public Hearing' a.k.a 'City Planning Commission Review'
       REVIEW_SESSION_PRE_HEARING_REVIEW_POST_REFERRAL, // aka 'Review Session - Pre-Hearing Review / Post Referral'
+      CPC_PUBLIC_MEETING_PUBLIC_HEARING, // aka 'CPC Public Meeting - Public Hearing' a.k.a 'City Planning Commission Review'
       REVIEW_SESSION_POST_HEARING_FOLLOW_UP_FUTURE_VOTES, // aka 'Review Session - Post Hearing Follow-Up / Future Votes'
+      CPC_PUBLIC_MEETING_VOTE, // aka 'City Planning Commission Vote'
```
godfreyyeung, line 188: Just re-organizing here to match business order

client/app/styles/modules/_m-actions-milestones.scss
```diff
@@ -14,18 +14,14 @@
   position: relative;
   box-shadow: inset rem-calc(28) 0 0 $body-background, inset rem-calc(30) 0 0 $medium-gray;
 
-  &.past {}
-
-  &.present {
+  &.green-glow {
```
godfreyyeung, line 17: Renaming this effect. It's not good to use CSS for logic, and renaming this variable to reflect the visual style
image

client/app/styles/modules/_m-actions-milestones.scss
```diff
@@ -56,41 +52,6 @@
   }
 }
 
-.milestone.present + .milestone.future {
```
godfreyyeung, line 59: Removing for now.
1) It wasn't stable because the assigning of `.present` and `.future` classes to milestone elements wasn't stable. (due to `tense()` cp instability)

2) we shouldn't rely on CSS to calculate the position of a "now" marker.

 In the future we can work out some component to insert a "now" marker.

client/app/templates/components/project-milestone.hbs
```diff
@@ -68,133 +71,131 @@
...
-{{yield}}
+                  </li>
+                {{/if}}
+              {{/each}}
+            {{/borough-pres-deduped-votes-list}}
+          </ul>
+        {{/if}}
+      {{/each}}
+    {{/hearings-list-for-milestones-list}}
+  {{/if}}
```
godfreyyeung, line 200: Aline mentioned that this other info (outcomes, links) should never show if the milestone is "not started" / until the milestone is "in progress" or "completed"

client/app/templates/components/project-milestone.hbs
```diff
@@ -27,31 +30,31 @@
...
-        {{else}}
-          <DateDisplay
-            @date={{milestone.displayDate}}
-            @outputFormat="MMMM D, YYYY"
-          />
-        {{/if}}
+  {{#if (and
+    (in-array (array "In Progress" "Completed") milestone.statuscode)
+    milestone.displayDate
+  )}}
```
godfreyyeung, line 36: Consolidating these two conditions. The date, outcome, link, hearing results, etc. should never show until the milestone is "In progress" or "completed" and there is a recorded date.

PR #1340, opened 2021-05-27
2792 eas display

allthesignals:
### Summary

Before, in the ZAP Search project page, the Environment Assessment Survey Milestone was using an inappropriate date. This fixes that by conditionally preferring a different milestone's date.

#### Tasks/Bug Numbers
 - Fixes [AB#2792](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/2792)

### Technical Explanation

It was using an inappropriate date field. This fix makes the milestone conditionally prefer a different milestone's date.

See the computed property in the milestone model, `displayDate` in the front end as well as the `transformMilestones` utility in the backend.

This begins to shift some of the milestone post-processing a little closer to the "frontend" side of the spectrum. Before, we were directly querying for specific list of milestone IDs. This was too heavy-handed because it only allowed for milestones that were to be displayed. So, if a milestone needed to be used for some value but _not_ displayed, this wouldn't work.

Instead, this change filters for "visible" milestones with a static list of milestone IDs. Later, the front-end filters for the "display" milestones. This way, the frontend can conditionally refer to non-display milestones (visible), as in the case of the EAS milestone date.

**List any ad-hoc, miscellaneous updates included**
added something to gitignore.

The gist of this ticket is that the "Environment Assessment Survey Milestone" needs to prefer a different display date in some cases. It was confusing at first because "Environment Assessment Survey Milestone" is not a milestone in CRM. This naming is instead an alias we introduce in the backend.


client/app/models/milestone.js
```diff
@@ -152,8 +153,16 @@ export default class MilestoneModel extends Model {
...
+    if (this.dcpMilestone === FILED_EAS_REVIEW) { // FILED_EAS_REVIEW is used here as the generic "EAS" milestone
+      // 2792: if the project starts with P, prefare the display date found in the "PREPARE_FIELD_EAS" milestone
+      if (this.project.get('dcpName').startsWith('P')) {
+        const prepareFiledEas = this.project.get('milestones').findBy('dcpMilestone', PREPARE_FILED_EAS);
+        if (prepareFiledEas) {
+          displayDate = prepareFiledEas.dcpActualenddate;
+        }
+      } else {
+        displayDate = this.dcpActualstartdate;
+      }
```
godfreyyeung, line 165: Amaazzing!!!!!! ⭐ 🌟  ⭐ 🌟  ⭐ 🌟  so clear!

godfreyyeung: So FILED_EAS_REVIEW is the same as "Review Filed EAS" in the ticket?

allthesignals: > So FILED_EAS_REVIEW is the same as "Review Filed EAS" in the ticket?

Yep! I could clean up the names sooner or later...

godfreyyeung (approved):
Thank you for this AMAZING PR. Seriously. It is clearer than the Azure ticket -- I read a similar one w/ @dhochbaum-dcp  earlier and didn't understand the reqs, but this ticket clarified for me! 

Thank you for the super thorough PR writeup. 

I left 1 question but approving in advance.

PR #1342, opened 2021-05-28
2794 update milestone display dates

dhochbaum-dcp:
### Summary
Updates some milestone display dates for In Progress and Completed items.

#### Tasks/Bug Numbers
 - Closes [AB#2794](https://dev.azure.com/NYCPlanning/ITD/_workitems/edit/2794)


client/app/models/milestone/constants.js
```diff
@@ -58,6 +62,7 @@ export const REVIEW_MILESTONE_IDS = [
 export const UPCOMING_MILESTONES = [
```
allthesignals, line 62: This UPCOMING_MILESTONES constant is references in MILESTONE_LIST_BY_TAB_LOOKUP, which is imported into the assignment section. So, this could affect that portion of the app, but it seems to be okay afaict.

client/app/models/milestone/constants.js
```diff
@@ -48,6 +48,10 @@ export const MAYORAL_VOTE = 'a83beec4-dad0-e711-8116-1458d04e2fb8';
 
 export const FINAL_LETTER_SENT = 'aa3beec4-dad0-e711-8116-1458d04e2fb8';
 
+export const REVIEW_FILED_LAND_USE_APPLICATION = '6c3beec4-dad0-e711-8116-1458d04e2fb8';
+
+export const ENVIRONMENTAL_ASSESSMENT_STATEMENT_FILED = '723beec4-dad0-e711-8116-1458d04e2fb8';
+
```
allthesignals, line 54: This is the main event — define these constants, import them in the milestone model, and add them to the groups in the computed property. Makes sense.

allthesignals (approved):

PR #1343, opened 2021-05-28
2795 2796 2797 city council public hearing

allthesignals:
### Summary

Show certain public hearing information below the City Planning Commission Review milestone. This information is derived from the milestone itself, rather than related dispositions.

#### Tasks/Bug Numbers
 - Fixes [AB#2795](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/2795)
 - Fixes [AB#2796](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/2796)
 - Fixes [AB#2797](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/2797)

### Technical Explanation
This is an enhancement rather than bug fix. The ask was to show some public hearing info.

The idiosyncratic issue was that this isn't a disposition-based hearing. Most of the time, hearings/recommendations are stored on the disposition. In this case, that information is stored on the milestone directly.

This mean some re-use of existing patterns but really needing to "duck type" some inputs. See the project-milestone.hbs diff. There, you'll find I've extracted some templating into a "public-hearing" component, and re-use it below, outside the context of the hearings/dispositions-deduplication work.

  c. Provide entrypoint of new solution, if different than (b).

project-milestone.hbs is the best starting point.

Consistent with existing patterns, for the most part...

One difference is that instead of checking for Milestone type in the template, I made a computed on the milestone directly.

The milestones/public-hearing component is just markup from project-milestone.hbs.

Separately, this appears to be a fix on an undefined reference:
image

Not sure if this is correct but it seems to have silently failed in the past, always leading to a "check" state.


client/app/models/disposition.js
```diff
@@ -91,12 +91,16 @@ export default class DispositionModel extends Model {
   // e.g. 'Favorable', 'Conditional Favorable', 'Unfavorable', 'Conditional Unfavorable',
   // 'No Objection', 'Waiver of Recommendation', N/A is defualt
 
+
+  // TODO: These are duplicates??? Are they numbers or strings?
```
godfreyyeung, line 95: what do you mean? one is borough **president*** the other is borough **board **

allthesignals: > what do you mean? one is borough **president*** the other is borough **board **

did you look at the other attrs on the model? 

image

client/app/models/disposition.js
```diff
@@ -91,12 +91,16 @@ export default class DispositionModel extends Model {
...
   @attr('number', { defaultValue: null }) dcpBoroughpresidentrecommendation;
 
   @attr('number', { defaultValue: null }) dcpBoroughboardrecommendation;
 
   @attr('number', { defaultValue: null }) dcpCommunityboardrecommendation;
 
+  @attr('number', { defaultValue: null }) dcpCitycouncilvote;
```
godfreyyeung, line 102: hmm I'm not seeing this field used anywhere ?

godfreyyeung (approved):

client/app/models/disposition.js
allthesignals: Ty, will remove. was vestigial

client/app/templates/components/project-milestone.hbs
```diff
@@ -195,6 +184,33 @@
...
   {{/if}}
+
+  {{#if (and
+    milestone.isCPCPublicMeetingPublicHearing
+    (is-same-or-after milestone.dcpReviewmeetingdate)
+  )}}
+    <ul class="sub-milestones no-bullet">
+      <li class="grid-x small-margin-bottom">
+        <div class="cell shrink small-margin-right">
+          {{#if (is-same-or-after milestone.dcpReviewmeetingdate)}}
```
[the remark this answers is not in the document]
allthesignals: yup. it's cribbed from some other markup that didn't have the same conditionals. will fix.

PR #1350, opened 2021-06-09
3215 Removed logic to add “Revised” to the beginning of milestones

dhochbaum-dcp:
### Summary
"Revised" no longer appears before any of the descriptions of duplicate milestones.

#### Tasks/Bug Numbers
 - Fixes [AB#3215](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/3215)

### Technical Explanation
~~I commented out the row that changes duplicate names.  Did not make more advanced changes to the logic as it was unnecessary and in case client wants to revert the change or make similar adjustments at a later date.~~

Revised to remove all logic for revising the name.


godfreyyeung (changes requested):
We should just remove `orderSensitiveName` because there's no longer a purpose. We should just display `displayName` 

We also need to remove the `isRevised` property

godfreyyeung (approved):
Thank you David!

PR #1353, opened 2021-06-14
Zap search odata query

jlhogan:
### Summary
Main task: replaces custom odata query building with this [odata-query](https://github.com/techniq/odata-query) builder package
Additionally:
  - removes XML query, using an additional OData query to the CRM to grab all necessary borough/block information for projects map geometries. This introduces some questions about how we want and need this to work. Currently pagination isn't implemented at all, and the CRM server query returns the same 500 results every time. Replacing the XML geom query also means we're limited to the block-level geometries of 500 projects. Given that the 500 result query is pretty performant in production, it might make sense to change the UX to do paginated queries of 500 results and show just those 500 results on the map as user paginates. Also possible the CRM max results could be upped to 5000? Otherwise, we can bring the XML querying back but things are much cleaner with out it. Either way, client <-> server pagination needs to be addressed _somehow_ in a follow up to this. Interested in thoughts @nnxka @allthesignals let's discuss next week?
  - removes any calls to CRM in geom service, to keep the limited to a projects-specific carto service
  - removes all references to crm `queryFromObject` except in the assignments service, which needs to happen in a fast follow
  - moves projects query and project detail query composition into utility helper functions
TODOS:
  - deal with pagination
  - remove `queryFromObject` from assignments query
  - clean up duplication of constants
  - possibly type-ify some of the nested project detail entities (probably not tho, p low reward)

*BLOCKERS*: not sure how to test the updates to non-project related routes or project sync route. Can someone help me come up with a validation plan?
  
#### Tasks/Bug Numbers
- [AB#3502](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/3502)


allthesignals:
Need to figure out why this request is failing... this is the interaction with Carto, and I'm not sure why it's having issues. Any thoughts?

@jlhogan 

image

allthesignals:
I think the `boroughBlocks`, which are now being retrieved from the OData endpoint, are no longer unique. Going to see if I Can make them unique in-memory...

rtblair:
@TylerMatteo please review.

TylerMatteo:
@rtblair I don't think this is really relevant to the issues around geometries that we were discussing. This PR moves a bunch of stuff into a new geometry "utility" from the service but I don't think it attempts to significantly change the actual code. Most of this is migrating to this `odata-query` package for using the OData. IMO I'm all for using a library like that instead of building our OData strings "by hand" but that's a huge undertaking and the way it's implemented here wouldn't necessarily be the approach I would take. At a high level, I don't see the point in doing any significant refactoring to either this API or the one in labs-applicant-portal in the foreseeable future. From what I can tell, they're based off the same code but copied into two apps for some reason and have diverged since. There's no reason zap search and applicant portal shouldn't use the same API (or set of APIs, depending on how they should be broken into microservices). That's a big project, obviously, but I don't think it's worth doing any big quality-of-life improvements to either API until we can undertake it.

PR #1354, opened 2021-06-14
3414 fix dispo statuscode and statecode

dhochbaum-dcp:
### Summary
Fixes issue where disposition statuscodes and statecodes were set to incorrect values on the front-end.

#### Tasks/Bug Numbers
 - Fixes [AB#3414](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/3414)
 
### Technical Explanation
Removes complex logic to figure out the statuscode and statecode which wasn't working, and simply uses the values that we already received from the server.

### DIFF
--- a/client/app/models/disposition.js
+++ b/client/app/models/disposition.js
@@ -121,50 +121,9 @@ export default class DispositionModel extends Model {
 
   @attr('string', { defaultValue: '' }) dcpBoroughpresidentrecommendation;
 
-  // sourced from statuscode
-  // Label: 'Draft', 'Value': 1
-  // Label: 'Saved', 'Value': 717170000
-  // Label: 'Submitted', 'Value': 2
-  // Label: 'Deactivated', 'Value': 717170001
-  // Label: 'Not Submitted', 'Value': 717170002
-  @attrComputed(
-    // recommendations
-    'dcpDateofvote',
-    'dcpBoroughpresidentrecommendation',
-    'dcpBoroughboardrecommendation',
-    'dcpCommunityboardrecommendation',
-    'dcpIspublichearingrequired',
+  @attr('string', { defaultValue: '' }) statuscode;
 
-    // hearings
-    'dcpPublichearinglocation',
-    'dcpDateofpublichearing',
-  )
-  get statuscode() {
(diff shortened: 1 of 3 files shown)


client/app/models/disposition.js
```diff
@@ -121,50 +121,9 @@ export default class DispositionModel extends Model {
...
-    'dcpBoroughpresidentrecommendation',
-    'dcpBoroughboardrecommendation',
-    'dcpCommunityboardrecommendation',
-    'dcpIspublichearingrequired',
-
-    // hearings
-    'dcpPublichearinglocation',
-    'dcpDateofpublichearing',
-  )
-  get statuscode() {
```
allthesignals, line 142: I'm pretty concerned about removing this logic because of how it might affect the LUP side of things

client/tests/unit/models/disposition-test.js
```diff
@@ -14,62 +12,6 @@ module('Unit | Model | disposition', function(hooks) {
     assert.ok(model);
   });
 
-  test('when participant opts out/submits, statuscode becomes "saved"/"submitted"', async function(assert) {
```
allthesignals, line 17: Hmm unless there's a good reason we should not be deleting tests in order to get the suite to pass. These are our only defense against regressions (read: grumpy emails).

allthesignals (changes requested):
@godfreyyeung can you take a look at this as well?

Let's huddle and figure out what the issue is. We want to preserve the LUP-side behavior while also addressing the issue you're trying to fix.

godfreyyeung:
> @godfreyyeung can you take a look at this as well?
> 
> Let's huddle and figure out what the issue is. We want to preserve the LUP-side behavior while also addressing the issue you're trying to fix.

We should meet with komal/liz/aline to see if it's still appropriate to derive statuscode and statecode on the frontend. I think it is better to read the value directly from the backend as it's less buggy. I don't remember context/pertinence of those frontend statuscode and statecode computed attributes.

dhochbaum-dcp:
We have agreed to move forward with using the data directly from CRM, as implemented above.

PR #1355, opened 2021-06-15
3453 download lup docs

allthesignals:
### Summary

Query for and display LUP's (a disposition record's) related documents, and enable download, according to certain filter criteria: 
image

#### Tasks/Bug Numbers
 - Fixes [AB#3453](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/3453)

 - Related [AB#3388](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/3388)

### Technical Explanation
Nothing super idiosyncratic here. This follows the same pattern for doc download/display as packages and artifacts.

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -83,6 +83,7 @@
   },
   "devDependencies": {
     "@nestjs/cli": "^7.6.0",
+    "@nestjs/schematics": "^7.3.1",
     "@types/express": "^4.17.1",
     "@types/jest": "^26.0.22",
     "@types/jsonwebtoken": "^8.3.4",
--- a/server/src/disposition/disposition.service.ts
+++ b/server/src/disposition/disposition.service.ts
@@ -1,0 +1,109 @@
+import { HttpException, HttpStatus, Injectable } from "@nestjs/common";
+import { SharepointService } from "../sharepoint/sharepoint.service";
+
+@Injectable()
+export class DispositionService {
+  constructor(private readonly sharepointService: SharepointService) {}
+
+  /**
+   * @param      {string}  relativeUrl    a string representing the Sharepoint URL relative to the entity folder.
+   *                                      In essence it is the Sharepoint folder name. e.g. P2015K0223_DOB Letter_3
+   * @param      {Object}  dcp_name   The disposition dcp_name property.
+   *                                  Used for error logging purposes.
+   *
+   * @return     {Object[]}     Array of 0 or more custom Document objects
+   */
+  async getdispositionSharepointDocuments(relativeUrl, dcp_name) {
+    if (relativeUrl) {
+      try {
+        const documents = await this.sharepointService.getSharepointFolderFiles(
+          `dcp_dispositions/${relativeUrl}`,
+          "?$expand=Files,Folders,Folders/Files,Folders/Folders/Files,Folders/Folders/Folders/Files"
+        );
+
+        if (documents) {
+          return documents.map(document => ({
+            name: document["Name"],
+            timeCreated: document["TimeCreated"],
+            serverRelativeUrl: document["ServerRelativeUrl"]
+          }));
+        }
+
+        return [];
+      } catch (e) {
+        if (e instanceof HttpException) {
+          throw e;
+        } else {
+          const errorMessage = `An error occured while constructing and looking up folder for disposition. Perhaps the disposition name or id is wrong. ${JSON.stringify(
+            e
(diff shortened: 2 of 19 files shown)


server/src/document/document.controller.ts
```diff
@@ -1,4 +1,5 @@
-import { Controller,
+import {
```
godfreyyeung, line 1: Are changes in this file all linting?

server/src/document/document.service.ts
```diff
@@ -118,59 +133,115 @@ export class DocumentService {
...
+    try {
+      const {
+        records: [firstDisposition]
+      } = await this.crmService.get(
+        "dcp_communityboarddispositions",
+        `
+        $select=dcp_communityboarddispositionid
+        &$filter=dcp_communityboarddispositionid eq ${recordId}
+          and (
+            dcp_visibility eq ${DISPOSITION_VISIBILITY.GENERAL_PUBLIC}
```
godfreyyeung, line 227: cool!

godfreyyeung: Thanks for all this research!

server/src/project/project.service.ts
```diff
@@ -507,6 +509,40 @@ export class ProjectService {
...
+    let projectDispositions =
+      transformedProject[
+        "dcp_dcp_project_dcp_communityboarddisposition_project"
+      ];
+
+    try {
+      projectDispositions = await Promise.all(
+        projectDispositions.map(async disposition => {
+          if (
+            [717170004, 717170003].includes(disposition.dcp_visibility) && // LUP or General Public
```
godfreyyeung, line 521: tiny tiny nit: can we use the consts here?

server/src/project/project.service.ts
```diff
@@ -620,7 +656,7 @@ export class ProjectService {
 
       dispositions: {
         ref: "dcp_communityboarddispositionid",
-        attributes: DISPOSITION_KEYS
```
godfreyyeung, line 623: why not follow same pattern as PACKAGE and ARTIFACTs, and place this key inside their constant lists?

server/src/project/project.service.ts
allthesignals: yep!

server/src/document/document.controller.ts
allthesignals: uh oh. yep except for the bottom of the file.

server/package.json
```diff
@@ -83,6 +83,7 @@
   },
   "devDependencies": {
     "@nestjs/cli": "^7.6.0",
+    "@nestjs/schematics": "^7.3.1",
```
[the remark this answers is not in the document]
allthesignals: yes

server/src/disposition/disposition.service.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * @param      {string}  relativeUrl    a string representing the Sharepoint URL relative to the entity folder.
+   *                                      In essence it is the Sharepoint folder name. e.g. P2015K0223_DOB Letter_3
+   * @param      {Object}  dcp_name   The disposition dcp_name property.
+   *                                  Used for error logging purposes.
+   *
+   * @return     {Object[]}     Array of 0 or more custom Document objects
+   */
+  async getdispositionSharepointDocuments(relativeUrl, dcp_name) {
```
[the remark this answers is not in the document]
allthesignals: oops ty

server/src/disposition/disposition.service.ts
```diff
@@ -0,0 +1,109 @@
...
+    if (relativeUrl) {
+      try {
+        const documents = await this.sharepointService.getSharepointFolderFiles(
+          `dcp_dispositions/${relativeUrl}`,
```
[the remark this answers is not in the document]
allthesignals: @godfreyyeung if it doesn't work (there are download links that don't download) I need to know about that

server/src/project/project.service.ts
allthesignals: Yeah, I think those other lists are probably fine but the problem is that they reference non-CRM attributes (like "documents") which we later learned in applicant portal was a problem for re-using them when building the CRM queries ($select=documents would give an error).

PR #1366, opened 2021-06-28
3236 - Added additional logic to existing functions to check to make sure we…

dhochbaum-dcp:
… want to show information.

#### Tasks/Bug Numbers
 - Fixes [AB#3236](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/3236)

### DIFF
--- a/client/app/components/hearings-list-for-milestones-list.js
+++ b/client/app/components/hearings-list-for-milestones-list.js
@@ -25,7 +25,7 @@ export function dedupeByParticipant(records = []) {
 // this is used to conditionally display the entire sub-milestone, including the title.
 export function checkHearingsSubmitted(records = []) {
   const projectsWithHearings = records.filter(function(disp) {
-    if ((disp.dcpDateofpublichearing !== null) && (disp.showHearingDetails)) {
+    if (disp.dcpDateofpublichearing !== null) {
       return disp.dcpPublichearinglocation && disp.dcpDateofpublichearing.toString();
     } return null;
   });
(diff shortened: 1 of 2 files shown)


godfreyyeung (changes requested):
Instead of hiding these two new checks into other booleans, you should make them explicit in the handlebars where the conditional occurs. 

For example, place them alongside the current conditional in `project-milestone.hbs` in line 69 and 93

client/app/components/hearings-list-for-milestones-list.js
```diff
@@ -44,6 +44,33 @@ export function checkVotesSubmitted(records = [], recommendationType) {
...
+      // null values and empty strings will NOT be returned
+      return disp.get(recommendationType);
+    }
+    return false;
+  });
+
+  return projectsWithVotes.length > 0;
+}
+
+
```
godfreyyeung, line 73: In the new functions, remove the logic already held by  checkVotesSubmitted and checkHearingsSubmitted. You can then either include all four booleans in the template, or have the two new funcitons depend on checkVotesSubmitted and checkHearingsSubmitted. For example:
export function checkHearingsShown(records = []) {
  const projectsWithHearingToShow = records.filter(function(disp) {
    if (disp.showHearingDetails) {
        return true
    } return false;
  });

       const areHearingsSubmitted = this.checkHearingsSubmitted(records);
  return areHearingsSubmitted && projectsWithHearingToShow
}

godfreyyeung (changes requested):

godfreyyeung (approved):

PR #1367, opened 2021-06-28
3625 downtime msg

allthesignals:
Now:
Screen Shot 2021-06-28 at 12 20 03 PM

When in maintenance mode:
c56031dc-34c5-45f9-adcf-7212faaeaa48

#### Tasks/Bug Numbers
 - Fixes [AB#3625](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/3625)

### DIFF
--- a/client/app/components/auth/login-dialog.js
+++ b/client/app/components/auth/login-dialog.js
@@ -1,6 +1,22 @@
 import Component from '@ember/component';
+import { computed } from '@ember/object';
 import ENV from 'labs-zap-search/config/environment';
+
+const MAINTENANCE_RANGE = ['06/28/2021 19:00', '06/29/2021 19:00'];
 
 export default Component.extend({
   oauthEndpoint: ENV.OAUTH_ENDPOINT,
+  isMaintenancePeriod: computed(function() {
+    const [from, to] = MAINTENANCE_RANGE.map(string => new Date(string));
+    const now = new Date();
+
+    return now > from && now < to;
+  }),
+
+  hasUpcomingMaintenance: computed(function() {
+    const [, to] = MAINTENANCE_RANGE.map(string => new Date(string));
+    const now = new Date();
+
+    return now < to;
+  }),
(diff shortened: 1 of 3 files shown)


client/app/templates/components/auth/login-dialog.hbs
```diff
@@ -14,9 +14,14 @@
...
 <p>Once you're signed in, you'll be redirected back here, where you'll see your list of assigned projects.</p>
 
+{{#if this.hasUpcomingMaintenance}}
+  <h1>Please note: This portal will be down for 24 hours of system maintenance starting Monday, June 28 at 7PM.</h1>
+{{/if}}
+
 <a
-  class="button large expanded"
-  href={{this.oauthEndpoint}}
+  class="button large expanded {{if this.isMaintenancePeriod 'disabled'}}"
```
godfreyyeung, line 22: Can we just hide the button instead?

client/app/components/auth/login-dialog.js
```diff
@@ -1,6 +1,22 @@
 import Component from '@ember/component';
+import { computed } from '@ember/object';
 import ENV from 'labs-zap-search/config/environment';
 
+const MAINTENANCE_RANGE = ['06/28/2021 19:00', '06/29/2021 19:00'];
```
godfreyyeung, line 5: What timezone is this? Should you use UTC timestamps instead? Does this default to UTC?

godfreyyeung (approved):

client/app/components/auth/login-dialog.js
allthesignals: probably. browsers defaults to the user's local timezone

client/app/templates/components/auth/login-dialog.hbs
allthesignals: we _can_

PR #1368, opened 2021-06-29
Promote develop to prod

allthesignals:


client/app/components/project-milestone.js
```diff
@@ -11,11 +11,7 @@ export default class ProjectMilestoneComponent extends Component {
...
   get getClassNames() {
-    let style = '';
-
-    if (this.milestone.statuscode === 'Not Started') style = 'gray';
-
-    if (this.milestone.statuscode === 'In Progress') style = 'green-glow';
+    const style = '';
 
     return `grid-x grid-padding-small milestone ${style}`;
   }
```
dhochbaum-dcp, line 17: Why set style at all?

dhochbaum-dcp (approved):

client/app/components/project-milestone.js
allthesignals: @dhochbaum-dcp not sure why it's like this. I probably has to do with the conditionals being removed. the whole computed should just be deleted and the classnames can be placed in-line.

PR #1395, opened 2022-05-02
8331 ZAP Search - Display Racial Equity Report ‘Applicability’ question to ZAP Search project page

dhochbaum-dcp:
### Summary
Display the new CRM “Applicability” question directly under the Status section on the ZAP Search project page.
Display the “Applicability” question as “Racial Equity Report Applicability”.
Display the Applicability question’s answer as = Null, Yes or No.
Display the following tooltip text when a user hoovers over the field: Identifies if applicant should submit a Racial Equity Report on Housing and Opportunity under Local Law 78 of 2021
Screen Shot 2022-05-02 at 2 39 01 PM

#### Tasks/Bug Numbers
 - Completes task [AB#8331](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/8331) and user story [AB#8123](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/8123)


client/app/models/project.js
```diff
@@ -120,6 +121,18 @@ export default class ProjectModel extends Model {
...
+  @attr() dcpApplicability;
+
+  @computed('dcpApplicability')
+  get dcpApplicabilitySimp() {
+    const option = Object.values(DCP_APPLICABILITY_OPTIONSET).find(x => x.code === this.dcpApplicability);
+    if (option) {
+      return option.label;
+    }
+
+    return 'Unknown';
```
horatiorosa, line 133: I wonder why we are returning `Unknown` instead of null. WIll that string value be saved to the sharepoint data?

dhochbaum-dcp: No, it won't be saved.  It is just in case the information is pulled up in the UI somewhere else, so what will be displayed as "Unknown" instead of "null".

horatiorosa: oh, yea, that makes sense. Ah, so the code that is saved would be from the `DCP_APPLICABILITY_OPTIONSET` const?

horatiorosa (approved):

client/app/models/project.js
```diff
@@ -120,6 +121,18 @@ export default class ProjectModel extends Model {
 
   @attr() videoLinks;
 
+  @attr() dcpApplicability;
+
+  @computed('dcpApplicability')
+  get dcpApplicabilitySimp() {
+    const option = Object.values(DCP_APPLICABILITY_OPTIONSET).find(x => x.code === this.dcpApplicability);
```
godfreyyeung, line 128: nit: Can you rename `x` to `option` or something ?

godfreyyeung: Just so the code is self documenting and we don't have single-letter variables

PR #1409, opened 2022-05-25
8729 - adjust NOTICED DATE to reflect EST.

horatiorosa:
### Summary
Adjusted for the time offset from GMT to EST so that the CRM and ZAP portal both displayed the same date as show in the screen grab below:

https://user-images.githubusercontent.com/11340947/170326647-78274ce7-f9a4-4167-965b-2018990a01d6.mov

#### Tasks/Bug Numbers
 - Fixes [AB#8729](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/8729)

### Technical Explanation
formatted `dcpNoticeddate` with a  computed function to reflect the offset from GMT to EST in the project js file

[Moment Usage](https://momentjs.com/docs/#/use-it/)

[UTC offset](https://momentjs.com/docs/#/manipulating/utc-offset/)

### DIFF
--- a/client/app/models/project.js
+++ b/client/app/models/project.js
@@ -1,6 +1,7 @@
 import DS from 'ember-data';
 import { computed } from '@ember/object';
 import { sort, alias } from '@ember/object/computed';
+import moment from 'moment';
 import { DCP_APPLICABILITY_OPTIONSET } from './project/constants';
 import {
   STATUSCODE_OPTIONSET,
@@ -124,6 +125,14 @@ export default class ProjectModel extends Model {
   @attr() dcpApplicability;
 
   @attr() dcpNoticeddate;
+
+  @computed('dcpNoticeddate')
+  get formattedNoticedDate() {
+    if (moment(this.dcpNoticeddate).isDST()) {
+      return moment(this.dcpNoticeddate).utcOffset(240).format('M/D/YYYY');
+    }
+    return moment(this.dcpNoticeddate).utcOffset(300).format('M/D/YYYY');
+  }
 
   @computed('dcpApplicability')
   get dcpApplicabilitySimp() {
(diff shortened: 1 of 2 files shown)


client/app/models/project.js
```diff
@@ -125,6 +126,11 @@ export default class ProjectModel extends Model {
 
   @attr() dcpNoticeddate;
 
+  @computed('formattedNoticedDate')
```
dhochbaum-dcp, line 129: This should be 'dcpNoticeddate', the variable used in the get function.

client/app/models/project.js
```diff
@@ -125,6 +126,11 @@ export default class ProjectModel extends Model {
...
+  get formattedNoticedDate() {
+    return moment(this.dcpNoticeddate).utcOffset(300).format('MM/D/YYYY');
```
dhochbaum-dcp, line 131: I believe that NYC is not always UTC -5, that it changes with daylight savings.  The docs you linked indicate that a more appropriate use in this case may be moment-timezone.

Also, why have leading zeroes on the month but not the day?  The site seems to use 'M/D/YYYY' pretty consistently.

dhochbaum-dcp (changes requested):

client/app/models/project.js
horatiorosa: @dhochbaum-dcp  great points. I thought it wouldn't matter since the only thing we were using the `dcpNoticeddate` was to display to users a date, since the difference or 4 or 5 hours would still get us the correct date. I refactored the computed method to check whether or not it is DST and adjust the appropriate offset amount. I tried to use `moment.tz(...)` but that wound up constantly breaking the app.

client/app/models/project.js
horatiorosa: noted and updated. thx!

PR #1414, opened 2022-06-29
Bug/1693 public meeting field overlaps bbl list

horatiorosa:
### Summary
If the light gray div containing the dcp action name and ULURP number is too long, it can possibly overlap and obscur the BBL numbers as shown in the example below:
[Project# 2020M0408](https://zap.planning.nyc.gov/projects/2020M0408)

Screen Shot 2022-06-29 at 11 11 34 AM

My solution, which I can't yet test locally bc I don't know how to recreate/duplicate the specific project locally, it to have the parent div wrapper display flex and center the child span. The results shown below. I am not sure of how this would visually affect shorter grey dics as yet.

Screen Shot 2022-06-29 at 5 11 54 PM

#### Tasks/Bug Numbers
 - Fixes [AB#1693](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/1693)

### DIFF
--- a/client/app/helpers/center-content.js
+++ b/client/app/helpers/center-content.js
@@ -1,0 +1,7 @@
+import { helper } from '@ember/component/helper';
+
+export function centerContent(text) {
+  return (text.length && text.offsetWidth > 200) ? 'center-text-tiny' : '';
+}
+
+export default helper(centerContent);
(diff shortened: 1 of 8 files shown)


godfreyyeung (approved):
Approving but have some suggestions for alternative approaches you can consider:
1) maxWidth on parent div, and extracting the text from spans to allow naturally breaking into new lines
2) if sticking with javascript, it will be more predictable to measure the div width.  alphanumerics vary in width, so 22 'i' is not the same width as 22 'W'

Looks like you have ember inspector already, so you can go into the "model" tab of that and modify the action text

horatiorosa:
> 

setting the CSS depending on the width is a good idea. I changed the helper to check for the `offSetWidth` of the text content.

PR #1418, opened 2022-07-05
Bug/1693 public meeting field overlaps bbl list 2

horatiorosa:
### Summary
I needed to move the `action.dcpName` within the each loops to access text length and add (or not) the class to center the ULURP name

#### Tasks/Bug Numbers
 - Fixes [AB#1693](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/1693)

### DIFF
--- a/client/app/templates/components/milestones/public-hearing.hbs
+++ b/client/app/templates/components/milestones/public-hearing.hbs
@@ -11,9 +11,12 @@
 <div class="text-tiny">
   {{#each @hearing.hearingActions as |action index|}}
     {{#if action.dcpName}}
-      <span class="label light-gray tiny-margin-bottom tiny-margin-right" data-test-hearing-actions-list="{{@hearing.disposition.id}}{{index}}">
-        {{action.dcpName}}
-        <small>{{action.dcpUlurpnumber}}</small>
+      <span class="{{if (is-text-too-lengthy action.dcpName) 'center-content'}}">
+        <span class="label light-gray tiny-margin-bottom tiny-margin-right"
+          data-test-hearing-actions-list="{{@hearing.disposition.id}}{{index}}">
+          {{action.dcpName}}
+          <small>{{action.dcpUlurpnumber}}</small>
+        </span>
       </span>
     {{/if}}
   {{/each}}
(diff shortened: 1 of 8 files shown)


client/app/templates/components/milestones/public-hearing.hbs
```diff
@@ -8,12 +8,14 @@
...
 </small>
 
-<div class="text-tiny" id="{{center-content action.dcpName}}">
+<div class="text-tiny">
   {{#each @hearing.hearingActions as |action index|}}
     {{#if action.dcpName}}
-      <span class="label light-gray tiny-margin-bottom tiny-margin-right" data-test-hearing-actions-list="{{@hearing.disposition.id}}{{index}}">
-        {{action.dcpName}}
-        <small>{{action.dcpUlurpnumber}}</small>
+      <span id="{{center-content action.dcpName}}">
```
godfreyyeung, line 14: `<span class="{{if (is-text-lengthy action.dcpName) 'center-content'}}">`
OR
`<span class="{{if (is-text-lengthy action.dcpName 22) 'center-content'}}">`

Few things here: 
1. the span should use `class`, not `id` because there will be multiple <span> tags in the page using the `center-content` style. 
2. To make the code more expressive and readable, the helper should be focused to determining whether a text string is over a certain length. 
Then, the template will use that helper to determine whether or not to render the center-content class.

`{{center-content acation.dcpName}}` indicates very little to a developer about its intentions. what does center-content return? why is action.dcpName passed into center-content? 

By renaming the helper to something like `is-text-lengthy`, we indicate the purpose is to determine the length of an input text. It likely returns a boolean. With this boolean, we determine how the span should be styled.

godfreyyeung (changes requested):

client/app/templates/components/milestones/public-hearing.hbs
horatiorosa: I renamed and refactored the helper method, and everywhere it was used (changed to a move obvious IF statement.

godfreyyeung (approved):

PR #1423, opened 2022-07-06
9499 use bracket notation to get array item, then get length of that item

horatiorosa:
### Summary
Turns out that the `action.dcpName` text is in an array with a length of 1 so the centerLengthyTextMethod will always return false and not apply the `center-content` class.

#### Tasks/Bug Numbers
 - Fixes [AB#9499](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/9499)

Screen Shot 2022-07-06 at 12 14 03 PM


bmarchena (approved):
looks good!

Issue #965, opened 2019-11-19
Stage: Same Project/Disposition displaying multiple time.


gopal2020:

gopal2020:
Project Name: Happy Halloween

Actual Result: Project name displaying multiple time on LUP Portal
Expected Result: Project name should displaying single time on LUP Portal.

![image](https://user-images.githubusercontent.com/57265341/69182317-bd1c7e00-0ade-11ea-9a0a-0c284f651aaf.png)

hannahkates:
This seems to occur when an LUP user is assigned to one project with the same role multiple times in CRM. I'm working on a SQL fix for the API that deduplicates the response when that happens.

gopal2020:
It is Fixed. Now displaying only one disposition.
![image](https://user-images.githubusercontent.com/57265341/69758400-6ee62b00-112d-11ea-99fb-493e9b521b9c.png)

ksinghprod:
This issue is back 

![image](https://user-images.githubusercontent.com/57362229/69982592-d3750180-1502-11ea-9719-67ce3a931867.png)

andycochran:
If this is because an LUP user is assigned to one project with the same role multiple times in CRM, then this is simply a data issue, not a bug.

ksinghprod:
@andycochran @hannahkates 

![image](https://user-images.githubusercontent.com/57362229/69991660-1be9ea80-1516-11ea-8ed5-cf31e3b7d152.png)

hannahkates:
This bug is resolved. QA Team please review.

ksinghprod:
This issue is not fixed, the view full project timeline option is not there 

![image](https://user-images.githubusercontent.com/57362229/70069099-35933c80-15bf-11ea-9d3b-f8c152c84f6b.png)

ksinghprod:
This issue is fixed, it may come after a deployment or patch

PR #1434, opened 2022-09-09
965 – Remove map tooltip on hover

bmarchena:
Resolves AB[#965](https://dev.azure.com/NYCPlanning/ITD/_workitems/edit/965) by removing the tooltip when hovering over points on the map. The tooltip previously displayed the project name and last milestone date by making references to columns on Carto (source layer: project-centroids, pic related) which appear to have since been removed.

Screen Shot 2022-09-09 at 9 03 15 AM

### DIFF
--- a/client/app/templates/components/projects-map-data.hbs
+++ b/client/app/templates/components/projects-map-data.hbs
@@ -11,11 +11,6 @@
   {{map.on 'click' (action 'handleMapClick')}}
   {{map.on 'mousemove' (action 'handleMouseMove')}}
 
-  {{#if highlightedFeature}}
-    <div class='map-tooltip' style="top:{{tooltipPoint.y}}px;left:{{tooltipPoint.x}}px;">
-      {{highlightedFeature.properties.dcp_projectname}} (<DateDisplay @date={{highlightedFeature.properties.dcpLastmilestonedate}} @outputFormat="MMMM, YYYY" />)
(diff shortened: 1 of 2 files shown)


client/app/templates/components/projects-map-data.hbs
```diff
@@ -11,11 +11,6 @@
   {{map.on 'click' (action 'handleMapClick')}}
   {{map.on 'mousemove' (action 'handleMouseMove')}}
 
-  {{#if highlightedFeature}}
-    <div class='map-tooltip' style="top:{{tooltipPoint.y}}px;left:{{tooltipPoint.x}}px;">
-      {{highlightedFeature.properties.dcp_projectname}} (<DateDisplay @date={{highlightedFeature.properties.dcpLastmilestonedate}} @outputFormat="MMMM, YYYY" />)
```
horatiorosa, line 16: It's a little disturbing that 2 columns were dropped from the dataset and not communicated to the OSE (or w/e the team was called when this happened) so we could at least update the codebase.

horatiorosa (approved):
🚀

client/app/templates/components/projects-map-data.hbs
bmarchena: agreed – these bugs will def keep cropping up unless we find a better way to communicate data changes (other than not at all)

PR #1473, opened 2023-02-15
Update links to ZR

bmarchena:
Previous links to ZR from ZAP were malformed. This PR adds a helper function to correctly construct new URLs to ZR.

### DIFF
--- a/client/app/helpers/build-url.js
+++ b/client/app/helpers/build-url.js
@@ -26,13 +26,41 @@ function zoningResolution(value) {
     AppendixJ: 'https://zr.planning.nyc.gov/appendix-j-designated-areas-within-manufacturing-districts',
   };
 
+  const romanNumerals = {
+    1: 'i',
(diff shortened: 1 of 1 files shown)


server/package.json
```diff
@@ -4,9 +4,6 @@
   "description": "",
   "author": "",
   "license": "MIT",
-  "engines": {
-    "node": "16.*"
```
TylerMatteo, line 8: I know there's some weirdness with getting the precommit hooks to work because the BE is on 16 and the FE is on 12 but unfortunately we can't just get rid of this. Does it work if you put this back and set it to `"16.* || 12.*"`?

TylerMatteo (changes requested):

server/package.json
bmarchena: It didn't work for me locally so I just added these lines back

TylerMatteo (approved):

PR #1480, opened 2023-05-16
13443 Exclude Non-Prod From Search Engine Indexing

dhochbaum-dcp:
### Summary

Goal is to ensure staging servers and other test deployments are not indexed by Google and other search engines.

#### Tasks/Bug Numbers
 - Completes AB#13443

### Technical Explanation

This could be done with the [ember-cli-noindex](https://github.com/elidupuis/ember-cli-noindex/blob/master/index.js) addon, but rather than add a dependency, I took the few lines of code that accomplish this and created an [Ember CLI In-Repo-Addon](https://cli.emberjs.com/release/writing-addons/in-repo-addons/).

### DIFF
--- a/client/lib/ose-noindex/index.js
+++ b/client/lib/ose-noindex/index.js
@@ -1,0 +1,27 @@
+'use strict';
+
+module.exports = {
+  name: 'ose-noindex',
+
+  /**
+   * Add `<meta name="robots" content="noindex, nofollow">` to <head>.
+   * We default to enabled in all environments except production.
+   */
+  contentFor(type, appConfig) {
+    const defaultConfig = {
+      enabled: appConfig.environment !== 'production',
+    };
+
+    const config = Object.assign({}, defaultConfig, appConfig.noindex || {});
+
+    if (config.enabled && type === 'head') {
+      return '<meta name="robots" content="noindex, nofollow" />';
+    }
+
+    return '';
+  },
+
+  isDevelopingAddon() {
+    return true;
+  },
+};
--- a/client/lib/ose-noindex/package.json
+++ b/client/lib/ose-noindex/package.json
@@ -1,0 +1,6 @@
+{
+  "name": "ose-noindex",
+  "keywords": [
+    "ember-addon"
+  ]
+}
--- a/client/package.json
+++ b/client/package.json
@@ -128,5 +128,10 @@
   },
   "lint-staged": {
     "*.js": "eslint --cache --fix"
+  },
+  "ember-addon": {
+    "paths": [
+      "lib/ose-noindex"
+    ]
   }
 }


client/lib/ose-noindex/index.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+const NYCID_CLIENT_ID = process.env.NYCID_CLIENT_ID || 'lup-portal-local';
```
TylerMatteo, line 2: We shouldn't use existing environment variables like `NYCID_CLIENT_ID` to infer environment and, at a higher level, I don't think using `process.env` at all is the idiomatic Ember way of doing this. If we changed the `netlify.toml` file so that environment wasn't always `production`, could you [borrow from how it's done in the source addon](https://github.com/elidupuis/ember-cli-noindex/blob/master/index.js#L15) so that you can set a flag in `environment.js`? I think that would let you just set a flag in the `if(environment === "production")` block of environment.js.

I think Ember wants you to set `environment` to "development" only when you're actually developing the app locally but using that value for non-production environments is something we do elsewhere already and is less hacky than the alternatives.

TylerMatteo (changes requested):

client/lib/ose-noindex/index.js
dhochbaum-dcp: I think the changes I made should work, but can't be sure that it will not add the line of code to production deploys.  If you think it looks good now, let me know when you want to deploy so we can double-check and change it back immediately if necessary.

client/lib/ose-noindex/index.js
```diff
@@ -0,0 +1,21 @@
...
+
+module.exports = {
+  name: 'ose-noindex',
+
+  /**
+   * Add `<meta name="robots" content="noindex, nofollow">` to <head>.
+   * We default to enabled in all environments except production.
+   */
+  contentFor(type, appConfig) {
+    if (!appConfig.IS_PROD && type === 'head') {
```
TylerMatteo, line 11: I don't want us to have duplicative `environment` and `IS_PROD` values. How about you copy the approach in the [source addon](https://github.com/elidupuis/ember-cli-noindex/blob/master/index.js#L15) by having an object on appConfig called `noindex` with a property called `enabled`. That way you can set it in environment.js with something like this.
```
if (environment === 'production') {
  ENV.noindex = {
    enabled: true
  }
}
```

TylerMatteo (changes requested):

client/lib/ose-noindex/index.js
dhochbaum-dcp: Ok, I have pushed changes that I believe are what you were requesting, but I am a little unclear as to how it is not just adding extra abstraction.

client/config/environment.js
```diff
@@ -148,6 +148,13 @@ module.exports = function(environment) {
     };
 
     ENV['mapbox-gl'].map.style = 'https://labs-layers-api.herokuapp.com/v1/base/style.json';
+    ENV.IS_PROD = true;
```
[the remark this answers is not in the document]
TylerMatteo: Otherwise, lgtm!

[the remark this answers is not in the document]
dhochbaum-dcp: Ok, removed that line.  We have only one  `if (environment !== 'production')` block.

PR #1504, opened 2024-04-03
1499 Map Shows Multiple Markers for a Single BBL

### LINKED ISSUE #1499: Map Shows Multiple Markers for a Single BBL
https://zap.planning.nyc.gov/projects?applied-filters=project_applicant_text&project_applicant_text=gompers
image
Note the two blue dots on the map.  They represent the same project, which has the associated BBL 2025760026.

The boro and block are used in the [geometry service](https://github.com/NYCPlanning/labs-zap-search/blob/develop/server/src/project/geometry/geometry.service.ts#L20) to query the cartodb table 'dtm_block_centroids_v20201106'.

But if we look at the [database in carto](https://nycplanning-web.carto.com/u/planninglabs/dataset/dtm_block_centroids_v20201106), we can see it is returning two centroids that match a boro=2 and block=2576.
image

While #1498 can be fixed without this issue being fixed, it is making it harder to troubleshoot #1498, and releasing a fix for #1498 without fixing this issue would undermine any confidence in its fuctionality.

dhochbaum-dcp:
I have created a [new table in Carto](https://nycplanning-web.carto.com/u/planninglabs/dataset/dof_dtm_block_centroids_deduped_by_boroblock) which has the contents of the `dof_dtm_block_centroids` table, with the following changes:
 - Any duplicate combinations of Boro and Block have been removed
 - `boro` has been renamed `borocode` to match the naming conventions of the previously used table

This fixes #1499, where the map showed multiple markers to represent the same block.

### DIFF
--- a/server/src/project/geometry/geometry.service.ts
+++ b/server/src/project/geometry/geometry.service.ts
@@ -18,7 +18,7 @@ const METERS_TO_FEET_FACTOR = 3.28084;
 const QUERIES = {
   DTM_BLOCK_CENTROIDS: `
     SELECT the_geom, the_geom_webmercator, cartodb_id, concat(borocode, LPAD(block::text, 5, '0')) as block 
-    FROM dtm_block_centroids_v20201106
+    FROM dof_dtm_block_centroids_deduped_by_boroblock
   `,
 
   centroidsFor(blocks) {


server/src/project/geometry/geometry.service.ts
```diff
@@ -18,7 +18,7 @@ const METERS_TO_FEET_FACTOR = 3.28084;
 const QUERIES = {
   DTM_BLOCK_CENTROIDS: `
     SELECT the_geom, the_geom_webmercator, cartodb_id, concat(borocode, LPAD(block::text, 5, '0')) as block 
-    FROM dtm_block_centroids_v20201106
+    FROM dof_dtm_block_centroids_deduped_by_boroblock
```
TylerMatteo, line 21: We'll have to come up with a solution that doesn't require us making a custom one-off table in Carto. The `dof_dtm_block_centroids` table is updated by GIS automatically and we don't want to introduce new table that we have to manually update. Moreover, I confirmed with Matt Croswell that block numbers aren't intended to be unique within a given borough.

TylerMatteo (changes requested):

server/src/project/geometry/geometry.service.ts
dhochbaum-dcp: I don't understand, if block numbers aren't unique within a given borough, wouldn't that mean that two lots could have the same BBL?

TylerMatteo: That was my first question but no. You can have two physical blocks with the same block number but they won't re-use lot numbers. You might have two blocks in one borough that are block "5", with three lots on each, but one would have lots 1, 2, 3 and the other would have lots 4, 5, 6, etc.

TylerMatteo: Honestly this might be rare enough that we don't have to worry about this bug unless it's actually blocking our ability to get radius search working. If the impact is just that users see two dots on the map for projects that happen to overlap with these duplicate blocks, it's not the worst thing.

dhochbaum-dcp: The issue with seeing two dots is that it undermines the credibility of the radius search.  When you draw a circle that contains two dots and only one result shows up, it just looks like the filter is broken anyway.

I tried and failed to figure out how to make the SQL only take 1 of each unique concatenated "block", which would solve the issue, but wasn't able to make it work.  That would fix the problem without having to create a new table.

Issue #1535, opened 2024-09-23
Add "Definitions of ZAP Updates" page to ZAP Search


TylerMatteo:
Add new "Definition of ZAP Updates" page to ZAP Search.

### Acceptance Criteria:
* New page should match [designs in Figma](https://www.figma.com/design/1HrOm1tkJ3lowQhrSRkPLs/ZAP-Listserv%3A-Wireframes-and-User-Stories?node-id=289-1197&node-type=frame&t=fNk1OU6QtQkJs14C-0)
* New page should be at route `/statuses`
* Link in "To sign up for ZAP updates, please visit here" should link to `/subscribe` (not implemented)
* Page should only be reachable if `SHOW_ALERTS` feature flag is `true`.

### Out of scope
* Adding "Email Notifications" button to banner at top of the site

@TylerMatteo to follow up with @jessicashanshanhuang about mobile designs

TylerMatteo:
@dhochbaum-dcp That link should link to `/subscribe`, as in the full URL in prod would be `zap.planning.nyc.gov/subscribe`.

dhochbaum-dcp:
Hi @jessicashanshanhuang, can you please take a look at https://deploy-preview-1537--labs-zap.netlify.app/statuses and let me know if the design is ok?

(The "subscribe" link is in the code, but since the page does not yet appear, the word linking it is not displaying)

jessicashanshanhuang:
Hey @dhochbaum-dcp, looks great overall! I just have some minor suggestions with the spacing: 

The spacing between the header and the "Definition of Zap Updates" should be increased, about 4rem.

![Image](https://github.com/user-attachments/assets/2b7f8ed7-d82a-40b5-b00d-738539fa1c96)

jessicashanshanhuang:
For this text:

![Image](https://github.com/user-attachments/assets/54b40623-2aa6-4e8e-a27c-ae1fc7b8ebf3)

Could we change the font size to 21px and have it bolded?

dhochbaum-dcp:
@jessicashanshanhuang I've made those updates, please re-check and let me know if you approve.

jessicashanshanhuang:
@dhochbaum-dcp Looks good to go  🫡

PR #1537, opened 2024-09-23
1535 Add "Definitions of ZAP Updates" page to ZAP Search

dhochbaum-dcp:
Add new "Definition of ZAP Updates" page to ZAP Search.

### Acceptance Criteria:

- [x] New page should match [designs in Figma](https://www.figma.com/design/1HrOm1tkJ3lowQhrSRkPLs/ZAP-Listserv%3A-Wireframes-and-User-Stories?node-id=289-1197&node-type=frame&t=fNk1OU6QtQkJs14C-0)
- [x] New page should be at route `/statuses`
- [x] Link in "To sign up for ZAP updates, please visit here" should link to `/subscribe` (not implemented)
- [x] Page should only be reachable if `SHOW_ALERTS` feature flag is `true`.

### Out of scope
* Adding "Email Notifications" button to banner at top of the site

@TylerMatteo to follow up with @jessicashanshanhuang about mobile designs

Note: The "subscribe" link is in the code, but since the route does not yet exist, the word linking to it is not displaying.

Completes #1535


dhochbaum-dcp:
@jessicashanshanhuang the statuses page can now be previewed at https://deploy-preview-1537--labs-zap.netlify.app/statuses

Please take a look and let me know if there are any changes you would like me to make.

client/app/router.js
```diff
@@ -28,6 +28,7 @@ Router.map(function() { // eslint-disable-line
   this.route('show-project', { path: '/projects/:id' });
   this.route('show-geography', { path: '/projects' });
   this.route('disclaimer');
+  config.showAlerts && this.route('statuses');
```
TylerMatteo, line 31: Can you fix this so we aren't breaking tests? You might just need to use an `if` instead of the `&&` shorthand.

TylerMatteo (changes requested):

client/app/router.js
dhochbaum-dcp: Sure thing

TylerMatteo (approved):

Issue #1536, opened 2024-09-23
Add ZAP Alerts feature flag to ZAP Search


TylerMatteo:
## Acceptance criteria
- [] A feature flag is created to toggle "New project" feature called `SHOW_ALERTS`
- [] Feature is opt-in (defaults to `false`)
- [] Feature is activated with Environment Variable

Based on feature flag work for Self Service in App portal. See corresponding [issue](https://github.com/NYCPlanning/labs-applicant-portal/issues/1242) and [PR](https://github.com/NYCPlanning/labs-applicant-portal/pull/1261/files#diff-6ee83be565750b9a226713abd37c1a340445e280c8bd60382f9017d4e298d45dR145) for reference.

TangoYankee:
Update PR to reference: https://github.com/NYCPlanning/labs-applicant-portal/pull/1263/files

PR #1539, opened 2024-09-26
Feature flag for zap alert

pratishta:
Closes https://github.com/NYCPlanning/labs-zap-search/issues/1536

### DIFF
--- a/client/config/environment.js
+++ b/client/config/environment.js
@@ -4,6 +4,7 @@ const { MIRAGE_SCENARIO } = process.env;
 const NYCID_CLIENT_ID = process.env.NYCID_CLIENT_ID || 'lup-portal-local';
 const NYC_ID_HOST =
   process.env.NYC_ID_HOST || 'https://accounts-nonprd.nyc.gov/account';
+const ZAP_ALERT = process.env.ZAP_ALERT || 'ON';
 
 module.exports = function (environment) {
   const ENV = {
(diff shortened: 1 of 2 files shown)


client/config/environment.js
```diff
@@ -17,6 +18,7 @@ module.exports = function (environment) {
     },
     NYC_ID_HOST,
     maintenanceTimes: getMaintenanceTimes(),
+    SHOW_ALERTS,
```
TangoYankee, line 21: My understanding is that the flag should resolve to a boolean value and it should be `false` by default.

From reading this, we take the environment string at face value without processing it into a boolean. When we use the value later in the code itself, we'll need to explicitly check its value every time

TangoYankee (changes requested):

client/config/environment.js
pratishta: Okay I updated it to use a function that returns a boolean depending on what it's set to like in the applicant portal PR. 

I notice we also have something similar in `config.services.ts` in applicant portal PR. Is `featureFlag` a more direct way of getting the feature flag env value when used in the server?

TangoYankee: > I notice we also have something similar in config.services.ts in applicant portal PR. Is featureFlag a more direct way of getting the feature flag env value when used in the server?

It is the way of using it in the server. Though, I think of it more in terms of "speaking the server's dialect" as opposed to "being more direct"

TylerMatteo (approved):

Issue #1544, opened 2024-10-07
New subscriber endpoint (part 1)


TylerMatteo:
Add new endpoint to ZAP Search API that creates a new contact in SendGrid and adds that user to the List for the current ZAP Search environment.

New endpoint should exist as a `POST` to `/subscribers` and take in a POST body with shape:
```
{
  "email": "john@example.com"
}
```

- [ ] Add SendGrid API key to ZAP Search back end as environment variable. API Key is availabe in SendGrid 1PW vault under "ZAP Search Staging SendGrid API Key"
- [ ] Add environment-specific List ID as environment variable

List names are `zap-subscribers-staging` for staging and `zap-subscribers-production` for production.

## Validation
The endpoint should perform the following validations on the supplied `POST` body and return the given error codes

- [ ] Confirm that provided email is a valid email address string. If invalid, return 400 Bad Request error response
- [ ] Confirm that supplied email address is not already a subscriber. Because we use SG in other projects, we can't rely on simply looking for a SG contact with the supplied email. Instead, a subscriber is considered to already exist if we have a contact with the given email **and** that email is already on the list for the current environment. If subscriber already exists, return [409 Conflict error response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409)

## Adding subscriber

- [ ] Given a call with valid data, a contact is added to SG and the contact is added to the correct List. Return `200` success response code.

Once data is validated, the subscriber should be created and added to the list. See SendGrid docs for adding/updating a contact [here](https://www.twilio.com/docs/sendgrid/api-reference/contacts/add-or-update-a-contact). Note that request body can take in List IDs that the contact should be added to.

Note the following from SG's docs describing how contact creation is an asynchronous process:

> Because the creation and update of contacts is an asynchronous process, the response will not contain immediate feedback on the processing of your upserted contacts. Rather, it will contain an HTTP 202 response indicating the contacts are queued for processing or an HTTP 4XX error containing validation errors. Should you wish to get the resulting contact's ID or confirm that your contacts have been updated or added, you can use the [Get Contacts by Identifiers operation](https://www.twilio.com/docs/sendgrid/api-reference/contacts/get-contacts-by-identifiers).

Our endpoint should still be synchronous, so endpoint code may require implementing polling with a max wait time. Let's start with 3 seconds and see how fast SG's APIs are.

TylerMatteo:
[...]

PR #1545, opened 2024-10-08
1544 New subscriber endpoint (part 1)

dhochbaum-dcp:
Add new endpoint to ZAP Search API that creates a new contact in SendGrid and adds that user to the List for the current ZAP Search environment.

New endpoint should exist as a `POST` to `/subscribers` and take in a POST body with shape:
```
{
  "email": "john@example.com"
}
```

- [ ] Add SendGrid API key to ZAP Search back end as environment variable. API Key is availabe in SendGrid 1PW vault under "ZAP Search Staging SendGrid API Key"
- [ ] Add environment-specific List ID as environment variable

List names are `zap-subscribers-staging` for staging and `zap-subscribers-production` for production.

## Validation
The endpoint should perform the following validations on the supplied `POST` body and return the given error codes

- [ ] Confirm that provided email is a valid email address string. If invalid, return 400 Bad Request error response
- [ ] Confirm that supplied email address is not already a subscriber. Because we use SG in other projects, we can't rely on simply looking for a SG contact with the supplied email. Instead, a subscriber is considered to already exist if we have a contact with the given email **and** that email is already on the list for the current environment. If subscriber already exists, return [409 Conflict error response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409)

## Adding subscriber

- [ ] Given a call with valid data, a contact is added to SG and the contact is added to the correct List. Return `200` success response code.

Once data is validated, the subscriber should be created and added to the list. See SendGrid docs for adding/updating a contact [here](https://www.twilio.com/docs/sendgrid/api-reference/contacts/add-or-update-a-contact). Note that request body can take in List IDs that the contact should be added to.

Note the following from SG's docs describing how contact creation is an asynchronous process:

> Because the creation and update of contacts is an asynchronous process, the response will not contain immediate feedback on the processing of your upserted contacts. Rather, it will contain an HTTP 202 response indicating the contacts are queued for processing or an HTTP 4XX error containing validation errors. Should you wish to get the resulting contact's ID or confirm that your contacts have been updated or added, you can use the [Get Contacts by Identifiers operation](https://www.twilio.com/docs/sendgrid/api-reference/contacts/get-contacts-by-identifiers).

Our endpoint should still be synchronous, so endpoint code may require implementing polling with a max wait time. Let's start with 3 seconds and see how fast SG's APIs are.

Completes #1544

### DIFF
--- a/server/src/subscriber/subscriber.controller.ts
+++ b/server/src/subscriber/subscriber.controller.ts
@@ -1,19 +1,11 @@
-import { Controller, Get, Post, Param, Query, Req, Res } from "@nestjs/common";
+import { Controller, Post, Req, Res } from "@nestjs/common";
 import { ConfigService } from "../config/config.service";
+import { SubscriberService } from "./subscriber.service";
 import { Request } from "express";
 import validateEmail from "../_utils/validate-email";
-import client from "@sendgrid/client";
-
-type HttpMethod = 'get'|'GET'|'post'|'POST'|'put'|'PUT'|'patch'|'PATCH'|'delete'|'DELETE';
 
 const PAUSE_BETWEEN_CHECKS = 3000;
 const CHECKS_BEFORE_FAIL = 10;
-
-function delay(milliseconds){
-  return new Promise(resolve => {
-    setTimeout(resolve, milliseconds);
-  });
-}
 
 @Controller()
 export class SubscriberController {
@@ -21,8 +13,8 @@ export class SubscriberController {
   list = "";
 
   constructor(
-    private readonly config: ConfigService
-   
+    private readonly config: ConfigService,
+    private readonly subscriberService: SubscriberService
   ) {
     this.apiKey = this.config.get("SENDGRID_API_KEY");
     this.list = this.config.get("SENDGRID_LIST");
@@ -30,32 +22,14 @@ export class SubscriberController {
 
   @Post("/subscribers")
   async subscribe(@Req() request: Request, @Res() response) {
-    const searchRequest = {
-      url: "/v3/marketing/contacts/search/emails",
-      method:<HttpMethod> 'POST',
-      body: {
-        "emails": [request.body.email]
-      }
-    }
-    const addRequest = {
-      url: "/v3/marketing/contacts",
-      method:<HttpMethod> 'PUT',
-      body: {
-        "list_ids": [this.list],
-        "contacts": [{"email": request.body.email}]
-      }
-    }
-      
     if (!validateEmail(request.body.email)) {
       response.status(400).send({
         error: "Invalid email address."
       })
       return;
     }
-    
-    client.setApiKey(this.apiKey);
 
-    const existingUser = await this.searchByEmail(request.body.email)
+    const existingUser = await this.subscriberService.findByEmail(request.body.email)
     if(![200, 404].includes(existingUser.code)) {
       console.error(existingUser.code, existingUser.message);
       response.status(existingUser.code).send({ error: existingUser.message })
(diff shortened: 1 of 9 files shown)


server/src/subscriber/subscriber.service.ts
```diff
@@ -0,0 +1,81 @@
...
+    // https://www.twilio.com/docs/sendgrid/api-reference/contacts/add-or-update-a-contact
+    try {
+      const result = await this.client.request(addRequest);
+      return {isError: false, result: result};
+    } catch(error) {
+      return {isError: true, ...error};
+    }
+  }
+
+  async checkCreate(email: string, @Res() response, counter: number = 0, checksBeforeFail: number, pauseBetweenChecks: number, list: string) {
```
TylerMatteo, line 60: Seeing that SG is taking much to long to finish creating the contact for this to be done "synchronously", let's switch gears here. Have our endpoint return successfully if the `PUT` to `/v3/marketing/contacts` is a success. This means you can remove all of the polling code because you're no longer waiting to see the result of creating the contact.

server/src/subscriber/subscriber.controller.ts
```diff
@@ -0,0 +1,80 @@
...
+      response.status(importConfirmation.code).send({
+        status: "error",
+        error: importConfirmation.message
+      })
+      return;
+    }
+
+    response.status(200).send({
+      status: "success",
+      user: importConfirmation[0].body.result[request.body.email]
```
TylerMatteo, line 75: Remove this `user` property from the response. I don't think we'll need it and we don't want to leak user metadata (including their subscriptions via custom properties) into responses unnecessarily.

TylerMatteo (changes requested):

server/src/subscriber/subscriber.controller.ts
dhochbaum-dcp: Done.

server/src/subscriber/subscriber.service.ts
dhochbaum-dcp: Removed it.

server/src/subscriber/subscriber.service.ts
```diff
@@ -0,0 +1,61 @@
+import { Injectable, Res } from "@nestjs/common";
+import { ConfigService } from "../config/config.service";
+import { Client } from "@sendgrid/client";
+const crypto = require('crypto');
```
TylerMatteo, line 4: Change this to `import crypto from 'crypto'`. You'll get a type error which can be resolved with `yarn add @types/node`

TylerMatteo, line 4: Change this to `import crypto from 'crypto'`. You'll get a type error which can be resolved with `yarn add @types/node`

TylerMatteo (changes requested):

server/src/subscriber/subscriber.service.ts
```diff
@@ -0,0 +1,61 @@
+import { Injectable, Res } from "@nestjs/common";
+import { ConfigService } from "../config/config.service";
+import { Client } from "@sendgrid/client";
+import crypto from 'crypto';
+
+
+type HttpMethod = 'get'|'GET'|'post'|'POST'|'put'|'PUT'|'patch'|'PATCH'|'delete'|'DELETE';
```
TylerMatteo, line 7: Note for posterity - It's not ideal to define this type here because it's copying a type from Sendgrid's package but they don't expect theirs and I couldn't find a better quick fix either.

TylerMatteo (approved):

server/src/subscriber/subscriber.service.ts
dhochbaum-dcp: Yep, I would have imported if possible.

Issue #1547, opened 2024-10-11
New subscriber endpoint (part 2)


TylerMatteo:
Add the following functionality to the endpoint built in #1544 

- [ ] Add a `subscriptions` object to the request body that can have properties for each of the 60 subscription topics.
    - [ ] The key for these properties should be the community district IDs found [here](https://github.com/NYCPlanning/labs-zap-search/blob/develop/client/app/helpers/lookup-community-district.js) - (`K01`, `K02`, etc). For citywide, use `CW`.
    - [ ] The value will be `1` to indicate a subscription
- [ ] Update the endpoint code to set the corresponding custom fields for the requested subscriptions to `1` on contact creation
- [ ] Set the "confirmed" custom field to `0` for the corresponding environment for the user on contact creation

### Example of the expected request body for a user subscribing to districts K01, M01, and citywide:
```
{
  "email": "test@example.com",
  "subscriptions": {
    "K01": 1,
    "M01": 1,
    "CW": 1
  }
}
```

TylerMatteo:
Dependent on #1560

PR #1562, opened 2024-10-21
1547 New subscriber endpoint (part 2)

dhochbaum-dcp:
Add the following functionality to the endpoint built in #1544 

- [x] Add a `subscriptions` object to the request body that can have properties for each of the 60 subscription topics.
    - [x] The key for these properties should be the community district IDs found [here](https://github.com/NYCPlanning/labs-zap-search/blob/develop/client/app/helpers/lookup-community-district.js) - (`K01`, `K02`, etc). For citywide, use `CW`.
    - [x] The value will be `1` to indicate a subscription
- [x] Update the endpoint code to set the corresponding custom fields for the requested subscriptions to `1` on contact creation
- [x] Set the "confirmed" custom field to `0` for the corresponding environment for the user on contact creation

### Example of the expected request body for a user subscribing to districts K01, M01, and citywide:
```
{
  "email": "test@example.com",
  "subscriptions": {
    "K01": 1,
    "M01": 1,
    "CW": 1
  }
}
```
- [x] Update code in the `/subscribe` endpoint to use the custom fields for subscriber ids instead of `anonymous_id`.
- [x] Custom fields are `zap_staging_id` for staging and `zap_production_id` for production.

Completes #1547 and #1568


server/src/_utils/validate-subscriptions.ts
TylerMatteo, line 1: This functionality feels specific enough to subscriptions that it should live in a file related to subscriptions directly. If we had a `repository` for subs I would say that but, since we don't it can go in the service. The list of valid subscription topics (`validCustomFieldNames`) can be a property on that class. The validation can be a private method on that class.

server/src/_utils/validate-subscriptions.ts
```diff
@@ -0,0 +1,24 @@
+const validCustomFieldNames = ["K01", "K02", "K03", "K04", "K05", "K06", "K07", "K08", "K09", "K10", "K11", "K12", "K13", "K14", "K15", "K16", "K17", "K18", "X01", "X02", "X03", "X04", "X05", "X06", "X07", "X08", "X09", "X10", "X11", "X12", "M01", "M02", "M03", "M04", "M05", "M06", "M07", "M08", "M09", "M10", "M11", "M12", "Q01", "Q02", "Q03", "Q04", "Q05", "Q06", "Q07", "Q08", "Q09", "Q10", "Q11", "Q12", "Q13", "Q14", "R01", "R02", "R03", "CW"];
```
TylerMatteo, line 1: Two things here:

* I think I prefer something like `topics` for the name of this, just so we aren't bleeding Sendgrid terminology in. But I know that approach is subjective so I'll leave it up to you
* One easy change we _can_ make regardless of naming is to add some typing. You can create a [string literal type](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types) out of this by added `as const` at the end:

```
const topics = ["K01","K02",...] as const
``` 
Then create a type for strings that should be one of these string
```
type Topic = typeof topics[number];
```

server/src/_utils/validate-subscriptions.ts
```diff
@@ -0,0 +1,24 @@
...
+const validCustomFieldValues = [1];
```
TylerMatteo, line 2: Same comments as for `validCustomFieldNames` apply here. I think this will eventually need `0` as a another option so the typing is still relevant.

server/src/_utils/validate-subscriptions.ts
```diff
@@ -0,0 +1,24 @@
...
+
+/**
```
TylerMatteo, line 4: +1 for function signature comments :D

server/src/subscriber/subscriber.controller.ts
```diff
@@ -26,6 +30,13 @@ export class SubscriberController {
       return;
     }
 
+    if(!validateSubscriptions(request.body.subscriptions)) {
```
TylerMatteo, line 33: This validation feels business logic-y so probably best to put it in the service. The controller calls `create` on the service, and that function handles validation.

TylerMatteo (changes requested):

server/src/_utils/validate-subscriptions.ts
dhochbaum-dcp: Would you prefer that, or would you prefer that I move it into a "_utils" folder in the subscriber folder?  That would match the way project-specific utils are in /project/_utils

server/src/_utils/validate-subscriptions.ts
dhochbaum-dcp: I don't think `topics` is a good name for this, because it doesn't really correspond to either the Sendgrid terminology nor ours.  I think `validCustomFieldNames` and `validCustomFieldValues` makes sense because any developer looking at this in the future would intuit that they are looking at a list of Community Board ids, and they correspond to custom field names in Sendgrid.

dhochbaum-dcp: > * One easy change we _can_ make regardless of naming is to add some typing. You can create a [string literal type](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types) out of this by added `as const` at the end:

I've made updates that I believe match what you're requesting, but I don't really understand the benefit of doing so.  You established that this is very specific to subscriptions, and so we aren't going to be exporting and using these types anywhere else.  Is there a benefit to adding all this code that gets compiled into the exact JS we had before?

server/src/_utils/validate-subscriptions.ts
TylerMatteo: I prefer it in the service. "Utils" folders tend to become unruly, especially when there are a bunch of them. IMO, they're one of those anti-patterns that most projects resort to sooner or later but this feels like a good chance to avoid leaning into it.

TylerMatteo (changes requested):

server/src/_utils/validate-subscriptions.ts
TylerMatteo: Sure, happy to explain. One benefit is simply developer ergonomics. If you have a variable or parameter set to this new type, and try to assign it to a string, you'll get auto-completion in VSCode to one of the possible values. It's a small thing but it's a nice benefit for not much work.

Another is how stricter typing gets you to think about possible error states you might otherwise not have accounted for. If you have a variable or parameter typed to this new type and you try to assign it to a regular old `string'. You might now know for sure that your string is one of the valid values, and typescript will yell at you for trying, which in turn might remind you to do validation, avoiding a possible bug. Again, its a small thing in this scenario but once you get used to thinking in types, it's not a ton of work and the benefits add up.

server/src/_utils/validate-subscriptions.ts
dhochbaum-dcp: > This functionality feels specific enough to subscriptions that it should live in a file related to subscriptions directly. If we had a `repository` for subs I would say that but, since we don't it can go in the service. The list of valid subscription topics (`validCustomFieldNames`) can be a property on that class. The validation can be a private method on that class.

Currently the validateSubscriptions function is run in the controller, so if the validation is a private method, the controller is unable to access it.

dhochbaum-dcp: @TylerMatteo I just pushed a branch which makes these changes the best I could with my understanding - please take a look, and if you'd like updates to it, perhaps we can pair.

TylerMatteo (approved):
Good job!

Issue #1564, opened 2024-10-24
Add logging for failed subscriber creation


TylerMatteo:

TylerMatteo:
@dhochbaum-dcp I don't have an exact polling rate and max time limit in mind but in general it can be longer with longer pauses between attempts - something like 5 minutes, checking every 10-30 seconds.

TylerMatteo:
@dhochbaum-dcp For context, I asked Sendgrid about webhooks to handle this sort of thing and they basically said to do what I described about:
![Image](https://github.com/user-attachments/assets/a73bcd68-2ae2-4a3c-8f85-d505060f4e8c)

TylerMatteo:
@dhochbaum-dcp - @horatiorosa did a [spike](https://github.com/NYCPlanning/ae-zoning-api/discussions/203) looking at different Sentry packages for Nest a while back because, at the time, Sentry didn't have an official one. However, they [came out with one recently](https://docs.sentry.io/platforms/javascript/guides/nestjs/) so I say let's try using that :D

PR #1565, opened 2024-10-24
1564 Add logging for failed subscriber creation

dhochbaum-dcp:
### Summary
Adds polling to check for completed subscriber creation, and logs any errors to Papertrail and Sentry.

#### Tasks/Bug Numbers
Completes #1564


server/src/subscriber/subscriber.controller.ts
```diff
@@ -53,6 +60,16 @@ export class SubscriberController {
...
     })
+
+    const errorInfo = {
+      email: request.body.email,
+      anonymous_id: addToQueue.anonymous_id,
+      lists: this.list
+    }
+
+    // Now we keep checking to make sure the import was successful
+    const importConfirmation = await this.subscriberService.checkCreate(addToQueue.result[1]["job_id"], response, 0, CHECKS_BEFORE_FAIL, PAUSE_BETWEEN_CHECKS, this.list, errorInfo);
```
TylerMatteo, line 71: Do we need this variable if we're not using it? This could flag as an unused variable in the linter.

server/src/subscriber/subscriber.service.ts
```diff
@@ -40,22 +41,68 @@ export class SubscriberService {
...
+    await delay(pauseBetweenChecks);
+
+    const confirmationRequest = {
+      url: `/v3/marketing/contacts/imports/${importId}`,
+      method:<HttpMethod> 'GET',
+    }
+
+    // https://www.twilio.com/docs/sendgrid/api-reference/contacts/import-contacts-status
+    try {
+      const user = await this.client.request(confirmationRequest);
```
TylerMatteo, line 92: Can you change this to `job` or something similar? This endpoint gives you back details about the request to import a contact, not an object representing an actual user, so this was confusing to me as someone coming to the code for the first time.

TylerMatteo (changes requested):
In addition to the comments, just take a look at removing the dependency on `@sentry/browser` as discussed

server/src/subscriber/subscriber.controller.ts
dhochbaum-dcp: It's fine, it will be used when we add the subscription confirmation email

dhochbaum-dcp:
@TylerMatteo Despite much effort, I have not been able to resolve "This branch cannot be rebased due to conflicts".  I don't think there's actually an issue.  Otherwise, this is ready for review.

TylerMatteo:
> @TylerMatteo Despite much effort, I have not been able to resolve "This branch cannot be rebased due to conflicts". I don't think there's actually an issue. Otherwise, this is ready for review.

It likely has something to do with the merge commit merging develop back onto this branch. I'll enable squash merging for now and, assuming it lets us do that, we can do that for this PR. I was going to suggest squashing at least some of these commits anyway. Long term, we should sit down and figure out how to do rebasing instead of merge commits with Git Tower because this problem will continue to come up.

TylerMatteo (approved):

dhochbaum-dcp:
> > @TylerMatteo Despite much effort, I have not been able to resolve "This branch cannot be rebased due to conflicts". I don't think there's actually an issue. Otherwise, this is ready for review.
> 
> It likely has something to do with the merge commit merging develop back onto this branch. I'll enable squash merging for now and, assuming it lets us do that, we can do that for this PR. I was going to suggest squashing at least some of these commits anyway. Long term, we should sit down and figure out how to do rebasing instead of merge commits with Git Tower because this problem will continue to come up.

I did rebase in Git Tower - no idea how it turned out like that.

Issue #1566, opened 2024-10-25
labs-zap-search


dhochbaum-dcp:

dhochbaum-dcp:
Will we need to pass the environment in the confirm link, along with the uuid?  Or is the plan for the API to search both the zap_staging_id and zap_production_id fields and use the result to determine which confirmation variable to change?

TylerMatteo:
You shouldn't need to pass the environment as another path param in the link, but the confirm link will link to the confirmation page for the corresponding environment, so `https://zap.planning.nyc.gov/subscribers/<uuid>/confirm` and `https://zap-staging.planninglabs.nyc/subscribers/<uuid>/confirm` for production and staging, respectively. 

Then when they visit the front end, it will call the backend for the corresponding environment which, in turn will know which custom field to use.

PR #1571, opened 2024-10-30
1566 Send confirmation email to new subscriber

dhochbaum-dcp:
### Summary
Send confirmation email to new subscribers when they are successfully signed up.

- [x] Once SendGrid contact import job is completed, send confirmation email to new subscriber
- [x] Email should follow [designs in Figma](https://www.figma.com/design/1HrOm1tkJ3lowQhrSRkPLs/ZAP-Listserv%3A-Wireframes-and-User-Stories?node-id=660-232&node-type=frame&t=anCCmmEHN6kABnCM-0)
- [x] Email content should list what user selected - All CDs grouped by borough as well as citywide
- [x] "Confirm" button should link back to page in ZAP Search for confirming subscription. Path will be `/subscribers/<uuid>/confirm`.
 
* Dependent on #1564 

#### Tasks/Bug Numbers
Completes #1566 

Email link needs to be updated once we have a URL for "Modify Subscriptions".

Code for the email is currently not in the repo, can be seen/edited here: https://mc.sendgrid.com/dynamic-templates/d-3684647ef2b242d8947b65b20497baa0/version/62d2fdea-926f-4443-970d-bf8c5f88474d/editor

### DIFF
--- a/server/src/subscriber/subscriber.service.ts
+++ b/server/src/subscriber/subscriber.service.ts
@@ -13,6 +13,8 @@ const validCustomFieldValues = [1] as const;
 export type CustomFieldValueTuple = typeof validCustomFieldValues;
 type CustomFieldValue = CustomFieldValueTuple[number];
 
+type ValidSubscriptionSet = {"K01": CustomFieldValue; "K02": CustomFieldValue; "K03": CustomFieldValue; "K04": CustomFieldValue; "K05": CustomFieldValue; "K06": CustomFieldValue; "K07": CustomFieldValue; "K08": CustomFieldValue; "K09": CustomFieldValue; "K10": CustomFieldValue; "K11": CustomFieldValue; "K12": CustomFieldValue; "K13": CustomFieldValue; "K14": CustomFieldValue; "K15": CustomFieldValue; "K16": CustomFieldValue; "K17": CustomFieldValue; "K18": CustomFieldValue; "X01": CustomFieldValue; "X02": CustomFieldValue; "X03": CustomFieldValue; "X04": CustomFieldValue; "X05": CustomFieldValue; "X06": CustomFieldValue; "X07": CustomFieldValue; "X08": CustomFieldValue; "X09": CustomFieldValue; "X10": CustomFieldValue; "X11": CustomFieldValue; "X12": CustomFieldValue; "M01": CustomFieldValue; "M02": CustomFieldValue; "M03": CustomFieldValue; "M04": CustomFieldValue; "M05": CustomFieldValue; "M06": CustomFieldValue; "M07": CustomFieldValue; "M08": CustomFieldValue; "M09": CustomFieldValue; "M10": CustomFieldValue; "M11": CustomFieldValue; "M12": CustomFieldValue; "Q01": CustomFieldValue; "Q02": CustomFieldValue; "Q03": CustomFieldValue; "Q04": CustomFieldValue; "Q05": CustomFieldValue; "Q06": CustomFieldValue; "Q07": CustomFieldValue; "Q08": CustomFieldValue; "Q09": CustomFieldValue; "Q10": CustomFieldValue; "Q11": CustomFieldValue; "Q12": CustomFieldValue; "Q13": CustomFieldValue; "Q14": CustomFieldValue; "R01": CustomFieldValue; "R02": CustomFieldValue; "R03": CustomFieldValue; "CW": CustomFieldValue}
+
 
 type HttpMethod = 'get'|'GET'|'post'|'POST'|'put'|'PUT'|'patch'|'PATCH'|'delete'|'DELETE';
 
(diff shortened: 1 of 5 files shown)


server/src/subscriber/subscriber.service.ts
```diff
@@ -183,5 +215,38 @@ export class SubscriberService {
     return validCustomFieldValues.includes(value as CustomFieldValue);
   }
 
+  /**
+   * Convert the uploaded subscriptions object into a format for Handlebars to use in the confirmation email
+   * @param {object} subscriptions - The set of CDs the user is subscribing to
+   * @returns {boolean}
+   */
+  private convertSubscriptionsToHandlebars(subscriptions: object) {
```
TylerMatteo, line 223: Try to make a type for this `subscriptions` argument that is stricter than just `object`.

server/src/subscriber/subscriber.service.ts
```diff
@@ -183,5 +215,38 @@ export class SubscriberService {
...
+    var handlebars = { "citywide": false, "boroughs": [] }
```
TylerMatteo, line 224: Would it make the templating in Sendgrid much more difficult if `boroughs` was an object wit keys for each borough instead of an array? I'm just worried having it as an array might be bug-prone.

TylerMatteo (changes requested):
@dhochbaum-dcp Looks pretty good overall. In addition to the two comments I left in code, I'd like you to try adding some unit tests for the methods in the subscriber service.

server/src/subscriber/subscriber.service.ts
dhochbaum-dcp: As far as I'm aware, it's not possible - I would have done it that way if it were, but I don't think you're able to iterate over anything that's not an array.

server/src/subscriber/subscriber.service.ts
```diff
@@ -12,6 +13,8 @@ const validCustomFieldValues = [1] as const;
...
...; "K13": CustomFieldValue; "K14": CustomFieldValue; "K15": CustomFieldValue; "K16": CustomFieldValue; "K17": CustomFieldValue; "K18": CustomFieldValue; "X01": CustomFieldValue; "X02": CustomFieldValue; "X03": CustomFieldValue; "X04": CustomFieldValue; "X05": CustomFieldValue; "X06": CustomFieldValue; "X07": CustomFieldValue; "X08": CustomFieldValue; "X09": CustomFieldValue; "X10": CustomFieldValue; "X11": CustomFieldValue; "X12": CustomFieldValue; "M01": CustomFieldValue; "M02": CustomFieldValue; "M03": CustomFieldValue; "M04": CustomFieldValue; "M05": CustomFieldValue; "M06": CustomFieldValue; "M07": CustomFieldValue; "M08": CustomFieldValue; "M09": CustomFieldValue; "M10": CustomFieldValue; "M11": CustomFieldValue; "M12": CustomFieldValue; "Q01": CustomFieldValue; "Q02": CustomFieldValue; "Q03": CustomFieldValue; "Q04": CustomFieldValue; "Q05": CustomFieldValue; "Q06": CustomFieldValue; "Q07": CustomFieldValue; "Q08": CustomFieldValue; "Q09": CustomFieldValue; "Q10": CustomFieldValue; "Q11": CustomFieldValue; "Q12": CustomFieldValue; "Q13": CustomFieldValue; "Q14": CustomFieldValue; "R01": CustomFieldValue; "R02": CustomFieldValue; "R03": CustomFieldValue; "CW": CustomFieldValue}
```
TylerMatteo, line 16: Ok I couldn't quite get this typing to work how I wanted either and the type guards you already have are good enough. This new `ValidSubscriptionsSet` type is a good improvement but I think we can avoid all of this `CustomFieldValue` copypasta. If you replace this with `type ValidSubscriptionSet = Record<CustomFieldName, CustomFieldValue>`, everything should still be happy. [Record](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) is a built in utility type that defines a type for an object with given types for it's keys and values - in this case `CustomFieldName` and `CustomFieldValue` respectively.

TylerMatteo (dismissed):

server/src/subscriber/subscriber.service.ts
TylerMatteo, line 16: Ok I couldn't quite get this typing to work how I wanted either and the type guards you already have are good enough. This new `ValidSubscriptionsSet` type is a good improvement but I think we can avoid all of this `CustomFieldValue` copypasta. If you replace this with `type ValidSubscriptionSet = Record<CustomFieldName, CustomFieldValue>`, everything should still be happy. [Record](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) is a built in utility type that defines a type for an object with given types for it's keys and values - in this case `CustomFieldName` and `CustomFieldValue` respectively.

TylerMatteo (changes requested):

TylerMatteo (approved):

Issue #1573, opened 2024-11-04
Get subscriber subscriptions endpoint


TylerMatteo:
- [ ] Create new `GET` endpoint at `/subscribers/<id>`
- [ ] Endpoint should look up subscriber with ZAP id custom property equal to given id
- [ ] Return 404 if user not found
- [ ] Return object with that user's current subscriptions:

```
{
  "K01": 1,
  "R01: 1,
}
```

PR #1574, opened 2024-11-04
1573 Get subscriber subscriptions endpoint

dhochbaum-dcp:
### Summary
- [x] Create new `GET` endpoint at `/subscribers/<id>`
- [x] Endpoint should look up subscriber with ZAP id custom property equal to given id
- [x] Return 404 if user not found
- [x] Return object with that user's current subscriptions:

```
{
  "K01": 1,
  "R01: 1,
}
```

#### Tasks/Bug Numbers
Completes #1573

### DIFF
--- a/server/src/subscriber/subscriber.service.ts
+++ b/server/src/subscriber/subscriber.service.ts
@@ -146,6 +146,39 @@ export class SubscriberService {
   }
 
   /**
+   * Fetch the user's list of subscriptions.
+   * @param {string} id - The user's zap_production_id or zap_staging_id.
+   * @returns {object}
+   */
+  async getSubscriptions(id: string) {
+    const query = `zap_production_id LIKE '${id}' OR zap_staging_id LIKE '${id}'`
+    const request = {
+      url: `/v3/marketing/contacts/search`,
+      method:<HttpMethod> 'POST',
+      body: { query }
+    }
+
+    // https://www.twilio.com/docs/sendgrid/api-reference/contacts/search-contacts
+    // https://www.twilio.com/docs/sendgrid/for-developers/sending-email/segmentation-query-language
+    try {
+      const subscriptions = await this.client.request(request);
+      if(subscriptions[0].body["contact_count"] === 0) {
+        return {isError: true, code: 404, message: "No users found."};
+      }
+      const environment = (subscriptions[0].body["result"][0]["custom_fields"]["zap_production_id"] === id) ? "production" : "staging";
+      var subscriptionList = {};
+      for (const [key, value] of Object.entries(subscriptions[0].body["result"][0]["custom_fields"])) {
+        if(key.startsWith(`zap_${environment}_`) && validCustomFieldNames.includes(key.replace(`zap_${environment}_`, "") as CustomFieldName)) {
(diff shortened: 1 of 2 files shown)


server/src/subscriber/subscriber.service.ts
```diff
@@ -145,6 +145,39 @@ export class SubscriberService {
     }
   }
 
+  /**
+   * Fetch the user's list of subscriptions.
+   * @param {string} id - The user's zap_production_id or zap_staging_id.
+   * @returns {object}
+   */
+  async getSubscriptions(id: string) {
+    const query = `zap_production_id LIKE '${id}' OR zap_staging_id LIKE '${id}'`
```
TylerMatteo, line 154: Let's look for the custom field for the current environment explicitly, instead of hard coding the field names and doing an `OR`. Could you use that sendgrid environment env variable you have in your other branch?

server/src/subscriber/subscriber.service.ts
```diff
@@ -145,6 +145,39 @@ export class SubscriberService {
...
+    }
+
+    // https://www.twilio.com/docs/sendgrid/api-reference/contacts/search-contacts
+    // https://www.twilio.com/docs/sendgrid/for-developers/sending-email/segmentation-query-language
+    try {
+      const subscriptions = await this.client.request(request);
+      if(subscriptions[0].body["contact_count"] === 0) {
+        return {isError: true, code: 404, message: "No users found."};
+      }
+      const environment = (subscriptions[0].body["result"][0]["custom_fields"]["zap_production_id"] === id) ? "production" : "staging";
```
TylerMatteo, line 168: Same as above - let's reference the environment explicitly instead of trying to determine it based on user data.

server/src/subscriber/subscriber.service.ts
```diff
@@ -178,6 +180,38 @@ export class SubscriberService {
         })
    }
 
+  /**
+   * Fetch the user's list of subscriptions.
+   * @param {string} id - The user's zap_production_id or zap_staging_id.
+   * @returns {object}
+   */
+  async getSubscriptions(id: string) {
+    const query = `${this.sendgridEnvironmentIdVariable} LIKE '${id}'`
```
TylerMatteo, line 189: Now that [Pratishta's PR](https://github.com/NYCPlanning/labs-zap-search/pull/1575) for the PATCH endpoint is merged in, can you rebase this branch onto develop and update this to use the `getUserById` method she added?

TylerMatteo (changes requested):

server/src/subscriber/subscriber.service.ts
dhochbaum-dcp: The getUserById method she added returns only the user's email.  Are you asking me to 
a) get rid of the getSubscriptions service, modify the getUserById method to return the subscriptions as well, and then modify both routes in the controller to handle the modified object returned from the service; or
b) leave them as two separate services?

TylerMatteo: Disregard, I misread what her method was doing.

TylerMatteo (approved):

Issue #1550, opened 2024-10-18
Confirm Subscriber Endpoint


TylerMatteo:
Create a `PATCH` endpoint in ZAP Search to update subscriber custom properties.

New endpoint should exist as `PATCH` to `/subscribers/<uuid>`. For now, request body should only expect one property that lets us set the user's "confirmed" property to `1`
```
{
    "confirmed": 1
}
```

- [ ] Validate the request body and attempt to update the "confirmed" property for a user to the given value:
- [ ] The UUID in the path params should be used to look up the user based on their `anonymous_id` property in SendGrid

See #1560 for details on what confirmed custom field will be in SendGrid. It should be `zap_staging_confirmed` for staging and `zap_production_confirmed` for production.

dhochbaum-dcp:
Is there a benefit to having user signup confirmation having the same endpoint as updating user subscriptions?

TylerMatteo:
> Is there a benefit to having user signup confirmation having the same endpoint as updating user subscriptions?

Yes - it's a best practice to create endpoints around "resources" (in this case `subscribers`) and the standard HTTP methods (POST, GET, PATCH, etc) whenever possible. Instead of having an endpoint specifically for "confirming" a subscriber, we can have an endpoint for the general task of editing a subscriber's information and it can be used for editing confirmation status as well as subscriptions.

PR #1575, opened 2024-11-07
1550 confirm patch endpoint

pratishta:
Closes https://github.com/NYCPlanning/labs-zap-search/issues/1550

I tested it with the following request body in Postman:

```
{ 
    "subscriptions": {
        "confirmed": 1
    }
}
```

and with a `zap_staging_id` pulled from Sendgrid contacts.

### DIFF
--- a/server/src/subscriber/subscriber.service.ts
+++ b/server/src/subscriber/subscriber.service.ts
@@ -5,11 +5,11 @@ import { MailService } from "@sendgrid/mail";
 import crypto from 'crypto';
 import * as Sentry from "@sentry/nestjs";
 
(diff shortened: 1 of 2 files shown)


server/src/subscriber/subscriber.service.ts
```diff
@@ -145,6 +145,66 @@ export class SubscriberService {
...
+
+  /**
+   * Update user subscription.
+   * @param {string} environment - The environment variable
+   * @param {string} email - The user's email
+   * @param {object} custom_fields - The users custom lists and preferences
+   * @returns {object}
+   */
+  async update(environment: string, email: string, custom_fields: object) {
+    var updated_custom_fields = Object.entries(custom_fields).reduce((acc, curr) => ({ ...acc, [`zap_${environment}_${curr[0]}`]: curr[1] }), {})
```
TylerMatteo, line 185: Can we stick with `let` or `const` and camelCased variable/parameter names?

TylerMatteo (changes requested):

server/src/subscriber/subscriber.service.ts
pratishta: Do you want me to update to `let` and camelCase for the `create` in here as well ?

Issue #1603, opened 2024-11-21
Endpoint to check if subscriber with email exists


TylerMatteo:
Create endpoint in ZAP Search API that can be used to check if there is a subscriber with a given email address.

- [ ] Create new `GET` endpoint at `/subscribers/email/<email>`
- [ ] Endpoint should look up Sendgrid contact with the given email and check if it belongs to the SG list for the current environment. If it does, return a `204` response with no response body
- [ ] If the given email is not a valid email address, return `400` response
- [ ] If email is valid:
   - [ ] If contact exists and is on the list, return `201` with a message that user is already subscribed
   - [ ] If contact exists and _confirmed = 0,  return `201` with a message that user is subscribed but needs to confirm
   - [ ] If contact exists and _confirmed = null, or contact is not found, return `404`.

PR #1604, opened 2024-11-26
1603/endpoint to check if user email exists

pratishta:
Closes https://github.com/NYCPlanning/labs-zap-search/issues/1603

### DIFF
--- a/server/src/subscriber/subscriber.controller.ts
+++ b/server/src/subscriber/subscriber.controller.ts
@@ -169,4 +169,21 @@ export class SubscriberController {
     });
     return;
   }
+
+  @Get("/subscribers/email/:email")
+  async getEmail(@Param() params, @Res() response) {
+    if (!validateEmail(params.email)) {
+      response.status(400).send({
+        error: "Invalid email address."
+      })
+      return;
+    } 
+
+
+    const existingUser = await this.subscriberService.findByEmail(params.email);
+    console.log(existingUser.code);
+    response.status(204).send();
+    // console.log(existingUser['1'].result['pyerakala@planning.nyc.gov'].contact.segment_ids);
+    return;
+  }
 }
--- a/server/src/subscriber/subscriber.controller.ts
+++ b/server/src/subscriber/subscriber.controller.ts
@@ -177,13 +177,18 @@ export class SubscriberController {
         error: "Invalid email address."
       })
       return;
-    } 
-
+    }
 
     const existingUser = await this.subscriberService.findByEmail(params.email);
-    console.log(existingUser.code);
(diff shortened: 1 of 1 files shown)


server/src/subscriber/subscriber.controller.ts
```diff
@@ -121,4 +121,26 @@ export class SubscriberController {
...
+    if (!validateEmail(params.email)) {
+      response.status(400).send({
+        error: "Invalid email address."
+      })
+      return;
+    }
+
+    const existingUser = await this.subscriberService.findByEmail(params.email);
+    const existingUserListId = `zap_${this.sendgridEnvironment}_id`;
+    if (existingUser.code === 404 || existingUser['1'].result[params.email].contact.custom_fields[existingUserListId]) {
```
TylerMatteo, line 136: Let's have this also check that the user is associated with the Sendgrid List.

TylerMatteo (changes requested):

server/src/subscriber/subscriber.controller.ts
pratishta: Forgot to negate the second condition

pratishta:
@TylerMatteo Ready for review again after incorporating new requirements with @dhochbaum-dcp

server/src/subscriber/subscriber.controller.ts
```diff
@@ -122,4 +122,44 @@ export class SubscriberController {
...
+        error: "User not found."
+      })
+      return;
+    }
+
+    const userExistsInCurrentList = existingUser['1'].result[params.email].contact.list_ids.includes(this.list);
+    const existingUserListConfirmed = existingUser['1'].result[params.email].contact.custom_fields[`zap_${this.sendgridEnvironment}_confirmed`];
+
+    if (userExistsInCurrentList && existingUserListConfirmed) {
+      response.status(201).send({
```
TylerMatteo, line 147: What about instead of returning a message that we'd have to do string comparison on in the front end, we return data about the user? (cc @dhochbaum-dcp )
```
{
  confirmed: true
}
```

TylerMatteo (changes requested):

server/src/subscriber/subscriber.controller.ts
pratishta: Can we do both? Like 
```
{
    confirmed: true
    message: "User already subscribed." 
}
```
so that we can be consistent with our API responses and having human readable language to fall back on ?

dhochbaum-dcp (approved):

dhochbaum-dcp (approved):

dhochbaum-dcp:
I love the robustness of the response object.

PR #1612, opened 2024-12-03
1609 Detect existing users and send modify subscriptions email

### LINKED ISSUE #1609: Detect existing users and send modify subscriptions email
On `/subscribe` page, detect emails of existing subscribers and prompt them to get email with link to modify subscriptions

### Dependent
* https://github.com/NYCPlanning/labs-zap-search/issues/1603
* https://github.com/NYCPlanning/labs-zap-search/issues/1610
* https://github.com/NYCPlanning/labs-zap-search/issues/1611

dhochbaum-dcp:
### Summary
This adds functionality to check if the user already exists.
- The code for the endpoint to check if the user exists is working, reliant on merging #1604 and #1621 when they are approved

#### Tasks/Bug Numbers
 - Closes #1609 

Emails to use for testing:
 - User subscribed but did not confirm - `dhochbaumdcp+20241121test1@gmail.com`
 - User subscribed and confirmed - `dhochbaumdcp+20241119test1@gmail.com`
 - User exists in Sendgrid but has never subscribed to ZAP updates for this environment -  `dhochbaumdcp+20241119test3@gmail.com`

### DIFF
--- a/client/app/controllers/subscribe.js
+++ b/client/app/controllers/subscribe.js
@@ -1,0 +1,71 @@
+import Controller from '@ember/controller';
+import { action } from '@ember/object';
+import fetch from 'fetch';
+import ENV from 'labs-zap-search/config/environment';
+
+const tlds = [".com", ".gov", ".edu", ".net", ".org"];
+
+export default class SubscribeController extends Controller {
+  lastEmailChecked = "";
+  emailAlreadyExists = false;
+  emailNeedsConfirmation = false;
+  startContinuouslyChecking = false;
+  emailSent = false;
+
+
+  @action
+  async checkExistingEmail(event) {
+    const email = event.target.value;
+    if(email === this.lastEmailChecked) { return; }
+    this.set('lastEmailChecked', email);
+    this.set('startContinuouslyChecking', true);
+
+    try {
+      const response = await fetch(`${ENV.host}/subscribers/email/${email}`);
+      const userData = await response.json();
+      
+      if (userData.error) {
+        this.set('emailAlreadyExists', false);
+        this.set('emailNeedsConfirmation', false);
+        this.set('emailSent', false);
+        return;
+      }
+
+      if (userData.message === "User already subscribed.") {
+        this.set('emailAlreadyExists', true);
+        this.set('emailNeedsConfirmation', false);
+      } else if (userData.message === "User is subscribed but must confirm.") {
+        this.set('emailNeedsConfirmation', true);
+        this.set('emailAlreadyExists', false);
+      }
+      return;
+    } catch (error) {
+      // We will receive an error if:
+      // a) the user does not exist in Sendgrid, or 
+      // b) their confirmed field is null.
+      // Either way, we don't need to log to console
+      this.set('emailAlreadyExists', false);
+      this.set('emailNeedsConfirmation', false);
+      this.set('emailSent', false);
+    }
+  }
+
+  @action
+  continuouslyCheckEmail(event) {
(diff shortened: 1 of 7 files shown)


client/app/controllers/subscribe.js
```diff
@@ -0,0 +1,92 @@
...
+      const userData = await response.json();
+
+      if (userData.error) {
+        this.set('emailAlreadyExists', false);
+        this.set('emailNeedsConfirmation', false);
+        this.set('emailSent', false);
+        return;
+      }
+
+      if (userData.message === 'User already subscribed.') {
```
TylerMatteo, line 38: I had Pratishta update the response from this API to include a boolean `confirmed` value. Let's use that instead of comparing against magic strings.

client/app/controllers/subscribe.js
```diff
@@ -0,0 +1,92 @@
+import Controller from '@ember/controller';
+import { action } from '@ember/object';
+import fetch from 'fetch';
+import ENV from 'labs-zap-search/config/environment';
+
+const tlds = ['.com', '.gov', '.edu', '.net', '.org'];
```
TylerMatteo, line 6: Instead of keeping track of our own list of tlds and using that as validation, can we use the same regex for emails already present in `subscription-form.js`? I'd rather we just check that the email address matches a _pattern_ than check against a list of tlds. If what they enter after `.` isn't an actual tld, they should get the same behavior as any other invalid email address.

client/app/controllers/subscribe.js
```diff
@@ -0,0 +1,92 @@
...
+      // Either way, we don't need to log to console
+      this.set('emailAlreadyExists', false);
+      this.set('emailNeedsConfirmation', false);
+      this.set('emailSent', false);
+    }
+  }
+
+  @action
+  continuouslyCheckEmail(event) {
+    if ((this.startContinuouslyChecking) || (tlds.includes(event.target.value.slice(-4)))) { this.checkExistingEmail(event); }
```
TylerMatteo, line 59: From testing this, it looks like it hits the API on every key press as soon as the user starts typing. This would be a good use case for implementing debouncing but that's maybe overkill. Instead, how about you validate the given value against the email regex and only call the API if it passes? That way we aren't spamming the API will calls that we know aren't going to find anything.

TylerMatteo (changes requested):

client/app/controllers/subscribe.js
dhochbaum-dcp: The API isn't being called on every key press as soon as the user starts typing.  API calls begin when the end of the email input is one of the TLDs, or when the user removes focus from the email input.  Only after this point will API calls occur after every keystroke.  Because most users will enter their email address and then not change the field, it will minimize the number of queries, while allowing for immediate feedback for those that do change the field (which is useful for testing/development).  Replacing the TLDs with the email regex and logic used in `subscription-form.js` would allow queries to begin after any 2 letters after the period, meaning almost all users would make two queries when they first enter their address.

Let me know if you still would like me to switch things up.

client/app/controllers/subscribe.js
dhochbaum-dcp: See conversation below.

client/app/controllers/subscribe.js
TylerMatteo: I'd still like to switch it out. I just think sticking with that regex is going to serve us better in the long run.

TylerMatteo (changes requested):

TylerMatteo (approved):

Issue #1610, opened 2024-12-02
Build endpoint to send user modify email


TylerMatteo:
Add endpoint that sends subscriber an email to modify their subscriptions

- [ ] New `POST` endpoint at `/subscribers/<id>/modify` (note no request body)
- [ ] Sends email using template from #1608 to user with given id

### Dependent
* #1608

dhochbaum-dcp:
This needs to be at `/subscribers/<email>/modify`.  Otherwise, we are leaking the id which we use for authentication to anyone who types in the email address.

pratishta:
@TylerMatteo Can you confirm? I vaguely remember this being addressed at sprint planning but not sure

PR #1613, opened 2024-12-04
1610/build endpoint to send user modify email

pratishta:
Closes https://github.com/NYCPlanning/labs-zap-search/issues/1610

### DIFF
--- a/server/src/subscriber/subscriber.controller.ts
+++ b/server/src/subscriber/subscriber.controller.ts
@@ -122,4 +122,13 @@ export class SubscriberController {
 
     response.send(updatedContact);
   }
+
+  @Get("/subscribers/:id/modify")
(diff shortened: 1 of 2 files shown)


server/src/subscriber/subscriber.controller.ts
```diff
@@ -122,4 +122,24 @@ export class SubscriberController {
...
+        error: "User not found."
+      })
+      return;
+    }
+
+    const userId = existingUser['1'].result[params.email].contact.custom_fields[`zap_${this.sendgridEnvironment}_id`];
+
+    // Send the modify email
+    await this.subscriberService.sendModifySubscriptionEmail(params.email, this.sendgridEnvironment, userId);
+    response.status(201).send({
```
pratishta, line 140: Sending this since it's a `POST` and also because the request hangs without a response. Not sure if there's a better way.

TylerMatteo (approved):

PR #1616, opened 2024-12-06
1555 Build endpoint for sending project updates

### LINKED ISSUE #1555: Build endpoint for sending project updates
Build new endpoint to trigger project update emails for given project

- [ ] New `POST` endpoint to `/projects/<projectid>/send-update`
- [ ] Calls Sendgrid APIs to create [single sends](https://www.twilio.com/docs/sendgrid/api-reference/single-sends/create-single-send) for project to correct CD and citywide lists with correct metadata.
- [ ] Endpoint requires `Authorization` header implementing `Basic` authorization method. Credentials should be `username:password` base64 encoded. Username can be `ae-admin`, password should be randomly generated UUID to be stored as environment variable. If given authorization header contains incorrect credentials, return `401 Unauthorized` [response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401).
- [ ] If single sends are created and scheduled succesfully, return `200`
- [ ] If project with given id not found, return `404`

dhochbaum-dcp:
### Summary
Build new endpoint to trigger project update emails for given project

- [x] New `POST` endpoint to `/projects/<projectid>/send-update`
- [x] Calls Sendgrid APIs to create [single sends](https://www.twilio.com/docs/sendgrid/api-reference/single-sends/create-single-send) for project to correct CD and citywide lists with correct metadata.
- [x] Endpoint requires `Authorization` header implementing `Basic` authorization method. Credentials should be `username:password` base64 encoded. Username can be `ae-admin`, password should be randomly generated UUID to be stored as environment variable. If given authorization header contains incorrect credentials, return `401 Unauthorized` [response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401).
- [x] If single sends are created and scheduled succesfully, return `200`
- [x] If project with given id not found, return `404`

#### Tasks/Bug Numbers
 - Closes #1555 

### Technical Explanation
I ended up not needing to add any dependencies for this endpoint. 🎉

We did not need anything to manage users, since we do not have multiple users.  `Basic` authorization only requires that we ensure that the `username:password` base64 encoded sent with the request is correct.  I added a module/service to handle this portion.  Because this is a different authentication than is used elsewhere on the server, and is limited to a single endpoint, I was able to put the authorization in the controller, rather than needing to add it as middleware.

We had discussed potentially using Handlebars, but there did not appear to be any packages that provide out-of-the-box support for our use case.  There is support for either using Handlebars as your primary means of templating, and support for using Handlebars to precompile your site using a CLI; but they are not suited for importing a Handlebars template, populating it on the fly, and returning an HTML string.  Because of this, it would be overkill for us to come up with a complex implementation when the MVP requires only a single email template for this purpose, when instead we could simply use template literals.


server/src/listserv/auth.module.ts
TylerMatteo, line 1: Let's avoid having multiple files with the same name. Since there's already an `auth.module.ts`, `auth.service.ts`, etc, call these something else. Having the file name match the name of the exported class is a good rule of thumb. So if the class is `ListservAuthModule`, you might have `listserv-auth.module.ts`. You could also consider replacing references to "Listserv" here with `Basic` since this is implementing the "Basic" authentication strategy.

server/src/send-update/send-update.controller.ts
```diff
@@ -0,0 +1,79 @@
...
+    private readonly listservAuthService: ListservAuthService
+  ) {
+    this.apiKey = this.config.get("SENDGRID_API_KEY");
+    this.list = this.config.get("SENDGRID_LIST");
+    this.sendgridEnvironment = this.config.get("SENDGRID_ENVIRONMENT")
+  }
+
+  @Post("/projects/:id/send-update")
+  async sendUpdate(@Param() params, @Req() request: Request, @Res() response) {
+    const validatedUser = await this.listservAuthService.validateUser(request.headers.authorization)
```
TylerMatteo, line 27: Nest recommends handling things like authentication with [Guards](https://docs.nestjs.com/guards). There's an example of that in [applicant portal](https://github.com/NYCPlanning/labs-applicant-portal/blob/main/server/src/authenticate.guard.ts) but you might find Nest's docs [here](https://docs.nestjs.com/security/authentication#implementing-the-authentication-guard) and [here](https://docs.nestjs.com/recipes/passport#extending-guards) more helpful. Note those examples use Passport but they should still be helpful even if you aren't using it. Those examples still have `AuthService` so the guard you add will likely use your service, not replace it. Once you have a Guard, you can add it to this route with a decorator instead of validated the call in the controller itself.

server/src/send-update/send-update.controller.ts
```diff
@@ -0,0 +1,79 @@
...
+      response.status(401).send();
+      return;
+    }
+
+    const project = await this.projectService.findOneByName(params.id);
+    // If no project is found, projectService returns HTTP error automatically, and this function does not continue
+
+    // Production names use underscores, staging use dashes
+    const segmentName = `zap${this.sendgridEnvironment === "production" ? "_production_" : "-staging-"}${project["data"]["attributes"]["dcp-borough"] === "Citywide" ? "CW" : project["data"]["attributes"]["dcp-validatedcommunitydistricts"]}`
+    const segmentIdResult = await this.sendUpdateService.getSegmentId(segmentName)
```
TylerMatteo, line 38: A project can be "citywide" and belong to multiple CDs so this function would have to identify the _list_ of segment IDs that the project corresponds to and create a single send for each.

server/src/send-update/send-update.controller.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  @Post("/projects/:id/send-update")
+  async sendUpdate(@Param() params, @Req() request: Request, @Res() response) {
+    const validatedUser = await this.listservAuthService.validateUser(request.headers.authorization)
+    if (!validatedUser) {
+      response.status(401).send();
+      return;
+    }
+
+    const project = await this.projectService.findOneByName(params.id);
```
TylerMatteo, line 33: Does this project data include the `dcp_visibility` property? If it does, I'd like to add check here that it is set to `General Public`. We'll check that in the Power Automate flow as well but it's good to have another check here as an extra barrier. Sorry I forgot to include this in the requirements.

TylerMatteo (changes requested):

server/src/send-update/send-update.controller.ts
dhochbaum-dcp: My mistake, I have a better sense of the location options now.  It seems that if a user is in multiple segments receiving a single send, they will only receive one copy of the email, so it makes more sense to create one single send that goes to all of the segments, than an individual send for each segment.

Also, I want to bring to your attention, it's possible for a project to be neither Citywide nor associated with any CD, in which case the project would not fall under any of our segments. Ex. https://zap.planning.nyc.gov/projects/2024Q0218

TylerMatteo: If a project belongs to several CDs, and a user is subbed to all of those CDs, we would expect them to receive multiple emails. I know it's not ideal but we've talked through it extensively with PS and they're ok with it. As for creating one send and using it for all relevant segments, that would be fine with me but I don't think it will work because the content of the email has to be dynamic based on the _segment_. For example, [the copy in the email templates mentions the CD](https://www.figma.com/design/1HrOm1tkJ3lowQhrSRkPLs/ZAP-Listserv%3A-Wireframes-and-User-Stories?node-id=190-28&node-type=frame&t=2Ojsd8uRqkYopgTn-0). 

Yeah, I figured there would be some that don't fall in any. That shouldn't be a concern.

server/src/send-update/send-update.controller.ts
TylerMatteo: I agree the guard is unlikely to get used more than once but I still think it's worth it - there's value in doing our best to stick to the framework's best practices and given the guard should be relatively simple to implement I think it's worth it. Plus, it'll be good experience for you with a framework we're planning to use more in the future. If you find yourself spending more than an afternoon getting the same functionality to work with a Guard, then we'll talk.

TylerMatteo (changes requested):

server/src/send-update/send-update.controller.ts
```diff
@@ -24,15 +27,9 @@ export class SendUpdateController {
...
-      response.status(401).send();
-      return;
-    }
-
     const project = await this.projectService.findOneByName(params.id);
     // If no project is found, projectService returns HTTP error automatically, and this function does not continue
-    
+
     // Production names use underscores, staging use dashes
     var segments = project["data"]["attributes"]["dcp-borough"] === "Citywide" ? [{ name: "CW", envSegment: `zap${this.sendgridEnvironment === "production" ? "_production_" : "-staging-"}CW`, segmentId: "" }] : [];
```
TylerMatteo, line 34: I just asked Pratishta to update the staging segments names to use underscores like production so go ahead and update this accordingly.

TylerMatteo (changes requested):

TylerMatteo (approved):

PR #1563, opened 2024-10-23
Pratishta/subscribe

pratishta:
Closes https://github.com/NYCPlanning/labs-zap-search/issues/1542

Issue #1559, opened 2024-10-18
Add informational modal to ZAP Search homepage


TylerMatteo:
- [ ] Add a modal that users see when they visit ZAP Search that directs them to the new subscriptions page
- [ ] Designs can be found [here](https://www.figma.com/design/1HrOm1tkJ3lowQhrSRkPLs/ZAP-Listserv%3A-Wireframes-and-User-Stories?node-id=353-74&node-type=frame&t=U8AKDO14tXxrlzv8-0) but we don't need to follow designs exactly if there is an existing modal component in ZAP Search we can use. However, copy should be as seen in designs
- [ ] User can dismiss the modal and it doesn't reappear when navigating between pages in the app
- [ ] Modal has "Do not show this message again" checkbox. If checked, the user should not see the message again when they visit the app again in the same browser.
- [ ] Link to `/subscribe" page

TylerMatteo:
Dependent on #1542

PR #1619, opened 2024-12-09
Add timeout and test environment app.import call

TylerMatteo:
These changes are intended to resolve two different errors we've been having with running tests for this repo.

1. The change to `testem.js` fix an error seen sporadically referring to a timeout being reached. This error would manifest with a reference to Testem not being loaded
2. The change to `ember-cli-build` is meant to fix the error seen since #1563 was merged in. As far as I can tell, this error was due to foundation not being imported when the app was run in `test` mode. This seems to have fixed that but not 100% sure.

Tests now execute successfully but fail. The error I'm seeing says "Assertion Failed: Attempted to register a view with an id already in use". My working theory is that this is due to use of Foundations modal component introduced in #1559, based on a [similar Issue](https://github.com/emberjs/ember.js/issues/13026) found in Ember's repo.

I'd like to merge this in to fix the existing issues, knowing the tests still fail, then have @dhochbaum-dcp investigate the new error.

### DIFF
--- a/client/app/components/info-modal.js
+++ b/client/app/components/info-modal.js
@@ -2,6 +2,7 @@ import Component from '@ember/component';
 import { action } from '@ember/object';
 
 export default class InfoModalComponent extends Component {
+  // We have manually overridden this to hide the modal by default when running tests in tests/index.html
   open = !window.localStorage.hideMessage;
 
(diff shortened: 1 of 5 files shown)


pratishta (approved):
lgtm

PR #1665, opened 2025-03-04
issue 1650 add advisory message bar for staging environment

### LINKED ISSUE #1650: Create a banner for ZAP Search staging environment to indicate they are meant for testing
- Create a banner to display - "Sandbox for Testing Purposes".
- Display the banner on all pages of the following ZAP Search and LUPP portal staging environments:
      - LUP and ZAP Search https://zap-staging.planninglabs.nyc/

Image

horatiorosa:
### Summary
Advisory message bar for staging environment
 - add advisory message component and template
 - added advisory message bar to application template
 - updated client environment with `getFeatureFlagShowSandboxWarning` function setting `FEATURE_FLAG_SHOW_SANDBOX_WARNING` env var

#### Tasks/Bug Numbers
 - Closes #1650


client/app/adapters/application.js
```diff
@@ -11,7 +11,6 @@ export default class ApplicationAdapter extends JSONAPIAdapter {
   session;
 
   get headers() {
-    console.log(this.session);
```
TylerMatteo, line 14: Remove this?

client/app/components/advisory-message-bar.js
```diff
@@ -0,0 +1,11 @@
+import Component from '@ember/component';
+import { tracked } from '@glimmer/tracking';
+import ENV from 'labs-zap-search/config/environment';
+
+export default class AdvisoryMessageBarComponent extends Component {
+  @tracked
+  showSandboxWarningOn = ENV.featureFlagShowSandboxWarning;
+
+  // @argument
+  message = false;
```
TylerMatteo, line 10: Maybe make this an empty string if you need a default value? That way you aren't mixing bools and strings on the same property.

client/app/templates/components/advisory-message-bar.hbs
```diff
@@ -0,0 +1,5 @@
+<div class="advisory-message">
```
TylerMatteo, line 1: Maybe check if `this.message` is truthy and conditionally render the `div` based on that? That way we never accidentally render an empty bar.

client/config/environment.js
```diff
@@ -18,6 +17,8 @@ module.exports = function (environment) {
     NYC_ID_HOST,
     maintenanceTimes: getMaintenanceTimes(),
     showAlerts: getFeatureFlagShowAlerts(),
+    featureFlagShowSandboxWarning: getFeatureFlagShowSandboxWarning(),
```
TylerMatteo, line 20: This is probably splitting hairs but do we consider this a "feature flag" _per say?_ @TangoYankee - thoughts?

TylerMatteo (dismissed):

client/config/environment.js
TangoYankee: It's a feature and we need to toggle it with environment flags :nerd_face:

client/app/adapters/application.js
horatiorosa: Yeah, it seemed unnecessary to have the session printing to the console. I was thinking this was probably left after some some debugging.

client/app/templates/components/advisory-message-bar.hbs
horatiorosa: I don't think we would accidentally render an empty div as there is a check in the application template looking for the truthy value of the feature flag.

Were you thinking of something like:
```
{{#if this.message}}
  <div class="advisory-message">
  // ... code
```

horatiorosa:
@TylerMatteo  👀

TangoYankee (approved):

Issue #1664, opened 2025-03-03
Invalid subscribe form submittable with enter key


TylerMatteo:
The subscribe form is submittable by pressing the `enter` key even when the button is disabled due to invalid form state.

- [ ] User can only submit form with enter key when button is enabled.

PR #1666, opened 2025-03-04
1664 Invalid subscribe form submittable with enter key

dhochbaum-dcp:
### Summary
The subscribe form is submittable by pressing the `enter` key even when the button is disabled due to invalid form state.

- [ ] User can only submit form with enter key when button is enabled.

#### Tasks/Bug Numbers
 - Completes #1664

### Technical Explanation
This solution disables submission on Enter completely.  It would be an extremely rare occurrence for a user to have a valid submission while in the email input field and attempt to submit with Enter, and they will still be able to submit using the button.  The action which validates whether the form data is valid for submission, as well as much of the form data, are stored in a sub-component, meaning allowing the form to submit on Enter only when valid would require significant refactoring.

### DIFF
--- a/client/app/controllers/subscribe.js
+++ b/client/app/controllers/subscribe.js
@@ -65,6 +65,11 @@ export default class SubscribeController extends Controller {
   }
 
   @action
+  stopSubscribeOnEnter(event) {
+    if (event.key === "Enter") { event.preventDefault(); }
+  }
+
+  @action
   continuouslyCheckEmail(event) {
     if ((this.startContinuouslyChecking) || (validateEmail(event.target.value))) { this.checkExistingEmail(event); }
   }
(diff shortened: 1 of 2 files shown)


TylerMatteo:
@dhochbaum-dcp maybe we just deprioritize this if we think a proper fix is a big lift. Disabling enter-to-submit is a pretty big a11y no-no, enough that I'm tempted to say leaving the bug in is preferable.

horatiorosa:
Couldn't we check if keydown is enter and the button is disabled to prevent accidental / incomplete form submission?

dhochbaum-dcp:
> Couldn't we check if keydown is enter and the button is disabled to prevent accidental / incomplete form submission?

The issue is that the action triggered by the event listener has to go in the Subscribe controller, while the `canBeSubmitted` action is in the SubscriptionForm component, and there's no way to pass that value up to the event listener.

PR #1679, opened 2025-03-19
Hide ZAP Alerts modal behind Feature Flag

### LINKED ISSUE #1678: Hide ZAP Alerts modal behind feature flag
# Acceptance Criteria

- [ ] Hide modal informing users of new ZAP Alerts functionality behind existing `SHOW_ALERTS` feature flag environment variable

TangoYankee:
Update info-modal component to hide when ZAP Alerts feature flag is turned off

closes #1678

### DIFF
--- a/client/app/components/info-modal.js
+++ b/client/app/components/info-modal.js
@@ -1,9 +1,10 @@
 import Component from '@ember/component';
 import { action } from '@ember/object';
+import ENV from 'labs-zap-search/config/environment';
 
 export default class InfoModalComponent extends Component {
   // We have manually overridden this to hide the modal by default when running tests in tests/index.html
-  open = !window.localStorage.hideMessage;
+  open = ENV.showAlerts === true && window.localStorage.hideMessage !== 'true';
 
   dontShowModalAgain = false;
 


client/app/components/info-modal.js
```diff
@@ -1,9 +1,10 @@
 import Component from '@ember/component';
 import { action } from '@ember/object';
+import ENV from 'labs-zap-search/config/environment';
 
 export default class InfoModalComponent extends Component {
   // We have manually overridden this to hide the modal by default when running tests in tests/index.html
-  open = !window.localStorage.hideMessage;
+  open = ENV.showAlerts === true && !window.localStorage.hideMessage;
```
horatiorosa, line 7: `!window.localStorage.hideMessage` should be changed to `window.localStorage.hideMessage === "true"`, should it not?

horatiorosa (changes requested):

client/app/components/info-modal.js
dhochbaum-dcp: No, `!window.localStorage.hideMessage` is correct.  `hideMessage` is set to `true` when a user checks the box that they no longer want to see the message, so `open` should be `true` when `hideMessage` is not set to `true`.

TangoYankee: Mmm yeah, I should also check that it doesn't mess up the tests mentioned in the comment

horatiorosa: @dhochbaum-dcp  I've heard from a reliable source that the browser cookies are [strings of key/values pairs](!window.localStorage.hideMessage). it seems to me that `!window.localStorage.hideMessage` only confirms there is a cookie called `hideMessage` but not that it's value is equal to `"true"`. I've set that cookie manually to `"foobar"` to avoid the flashing modal bug.

TangoYankee: > !window.localStorage.hideMessage wouldn't return anything. To work with the local storage objects, we need to use the [Web Storage API](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage#specifications) methods (eg getItem, setItem, removeItem, clear). Nothing else would work.

I'm not sure how you're testing each method. However, I'm using the in-browser console. I found that when the `hideMessage` key is not present in localStorage the `localStorage.hideMessage` syntax returns `undefined` and the `localStorage.getItem('hideMessage')` syntax returns `null`. When the `hideMessage` key is present, both syntaxes return `"true"`

Consequently, both syntaxes are functional. However, they are not interchangable. Also, for the sake of keep the ticket scoped to the original concern, I'd like to keep the existing syntax.

TangoYankee: > @dhochbaum-dcp I've heard from a reliable source that the browser cookies are [strings of key/values pairs](!window.localStorage.hideMessage). it seems to me that `!window.localStorage.hideMessage` only confirms there is a cookie called `hideMessage` but not that it's value is equal to `"true"`. I've set that cookie manually to `"foobar"` to avoid the flashing modal bug.

To clarify, the code suggestion was missing the `!` and this does flip the check. When implementing the suggestion, I did keep it:

```js
open = ENV.showAlerts === true && window.localStorage.hideMessage !== 'true'
```

horatiorosa (approved):
🚢

PR #1704, opened 2025-09-22
Improve error handling and logging for artifactWithDocuments method in artifact service

### LINKED ISSUE #1703: Add additional logging for ZAP Search document loading errors
Sarah Whitham reported users are experiencing longer than normal load times when viewing projects in ZAP Search and are intermittently unable to download documents from projects when they do load. She reported the issue for [projects 2024Q0304](https://zap.planning.nyc.gov/projects/2024Q0304) but said the users have had these issues for other projects.

* Because these problems happen intermittently, we haven't been able to identify a root cause or isolate the issue to AE code or something upstream.
* One lead from our logs is occurences of these ["Error loading documents for artifact" errors](https://github.com/NYCPlanning/labs-zap-search/blob/master/server/src/artifact/artifact.service.ts#L91) ([logs link](https://my.na-01.cloud.solarwinds.com/210372719448513536/logs?focus=1901702091980455958&q=%22Error%20loading%20documents%20for%20artifact%22))

To help us debug these issues should they reoccur, we want to do the following:
- [ ] Add additional informational and error logging for the code path that hits this error message
- [ ] Ensure try/catch blocks in that code path log errors as given instead of overriding with custom error messages

TylerMatteo:
### Summary
This PR improves the error handling and logging for the `artifactWithDocuments` method in `artifact.service.ts`. We noticed that the `e` being caught by the try/catch wasn't actually being logged or used. Instead, that catch block also hard coded `errorMessage` and threw a new `404` `HttpException`

#### Tasks/Bug Numbers
 - Closes #1703 

### Technical Explanation
a. I added a new `if (e instanceof HttpException)` block in the try catch. This is meant to mirror the pattern seen in methods downstream of this one such as `getArtifactSharepointDocuments` and `sharepointService.getSharepointFolderFiles`. This should mean that errors thrown by downstream methods are successfully bubbled up instead of being overwritten by `artifactWithDocuments`
b. The code that now lives in the `else` block of that if now logs out `e` instead of the hard coded `errorMessage`. This should ensure any error caught here that _isn't_ and instance of `HttpException` is simply logged as is for maximum transparency. The hard coded `errorMessage` is set as the `detail` property on the 404 `HttpException` that this catch block throws so it's still being logged.

### DIFF
--- a/server/src/artifact/artifact.service.ts
+++ b/server/src/artifact/artifact.service.ts
@@ -88,17 +88,21 @@ export class ArtifactService {
           )
         };
       } catch (e) {
-        const errorMessage = `Error loading documents for artifact ${dcp_name}.`;
-        console.log(errorMessage);
+        if (e instanceof HttpException) {
+          throw e;
+        } else {
+          const errorMessage = `Error loading documents for artifact ${dcp_name}.`;
+          console.log("Error Message:", e);
 
-        throw new HttpException(
-          {
-            code: "ARTIFACT_WITH_DOCUMENTS",
-            title: "Artifact Documents Error",
-            detail: errorMessage
-          },
-          HttpStatus.NOT_FOUND
-        );
+          throw new HttpException(
+            {
+              code: "ARTIFACT_WITH_DOCUMENTS",
+              title: "Artifact Documents Error",
+              detail: errorMessage
+            },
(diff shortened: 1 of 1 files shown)


TangoYankee (approved):

TangoYankee:
:fire: :potato:

